PR #9, opened 2023-03-10
6 add ci workflows into the GitHub directory

heevasti:


.github/workflows/github-ci.yml
```diff
@@ -0,0 +1,119 @@
...
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python-latest
+        uses: actions/setup-python@v4
+        with:
+          # Take latest Python 3 version
+          python-version: "3.x"
```
rbudhrani, line 31: currently QMI is officially supported for Python 3.8 onwards so we should also run these steps with the older versions

.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,48 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
+# GitHub recommends pinning actions to a commit SHA.
+# To get a newer version, you will need to update the SHA.
+# You can also reference a tag or branch, but the action may change without warning.
+
+name: Publish to Pypi
```
rbudhrani, line 10: `Pypi` to `PyPI`

.github/workflows/schedule-full-ci-test-main.yml
```diff
@@ -0,0 +1,153 @@
...
+        uses: actions/upload-artifact@v3
+        with:
+          name: pylint-badge
+          path: pylint.svg
+        if: ${{ always() }}
+      - name: mypy
+        run: |
+          pip install mypy
+          mypy --namespace-packages $SOURCE_DIRS | tee mypy.log
+          if [ -n "$( tail -n 1 mypy.log | grep -e '^Succes' )" ]; then RESULT="passed"; else RESULT="failed"; fi
```
rbudhrani, line 69: `Succes` to `Success`

tests/instruments/picoquant/test_library_wrapper.py
```diff
@@ -1,17 +1,15 @@
 import unittest.mock
```
rbudhrani, line 1: why did we have to remove some tests?

rbudhrani (changes requested):

.github/workflows/github-ci.yml
heevasti: I was thinking of leaving that for the "daily" checker only in https://github.com/QuTech-Delft/QMI/blob/6-add-ci-workflows-into-the-github-directory/.github/workflows/schedule-full-ci-test-main.yml. Like we have now. This to reduce the total run-time of a simple "push" action.

.github/workflows/schedule-full-ci-test-main.yml
heevasti: This matching works adequately like this so no need to add the extra 's'.

.github/workflows/python-publish.yml
heevasti: To be fixed.

tests/instruments/picoquant/test_library_wrapper.py
heevasti: License reasons. I will put the ones we can test later on, when I have the necessary licensing text in.

.github/workflows/github-ci.yml
rbudhrani: this pipeline also runs for a PR, so we should check for the different versions before merging, especially since we see outside contributors making PRs in the future. Also the master branch should be protected so now one can push to it directly

heevasti: OK, done. Added a separate workflow file for pull requests.

rbudhrani (approved):

PR #14, opened 2023-03-21
13 modify workflows to do only python latest on push branch or pull request and full ci when doing a merge

rbudhrani:


.github/workflows/pull_request.yml
heevasti, line 1: Part of this ticket was also to make this workflow to run only when the pull request is, either:
- ready for review (like now) OR
- auto merge is enabled

See the specific action options in https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request. Now it will run each time when a push is made to this pull request. I think the best would be when the PR is made ready for review, or enabling possibly both options.

README.md
```diff
@@ -1,8 +1,8 @@
-![pylint](https://github.com/QuTech-Delft/QMI/tree/main/.github/badges/pylint.svg)
- ![mypy](https://github.com/QuTech-Delft/QMI/tree/main/.github/badges/mypy.svg)
- ![radon-cc](https://github.com/QuTech-Delft/QMI/tree/main/.github/badges/radon-cc.svg)
- ![radon-mi](https://github.com/QuTech-Delft/QMI/tree/main/.github/badges/radon-mi.svg)
- ![coverage](https://github.com/QuTech-Delft/QMI/tree/main/.github/badges/coverage.svg)
+![pylint](.github/badges/pylint.svg?sanitize=true)
+ ![mypy](.github/badges/mypy.svg)
```
heevasti, line 2: Won't these other files also eventually need this "sanitize" option? Or is this still part of the test, regarding to the fact that this is really done only at the pull request level?

.github/workflows/pull_request.yml
heevasti: I had also left these "allowed to fail" options in the workflow for testing. These could be removed now (note that mypy raises an error but it passes due to this option now). The mypy error should be fixed as in gitlab.

README.md
rbudhrani: ah yes, I was trying that out, but forgot to remove it at the end. Eventually using relative paths is what made it work and not sanitize. I'll remove the sanitize option

.github/workflows/pull_request.yml
rbudhrani: if we triggered the workflow to run on `ready_for_review` or `auto_merge_enabled` then pushing anything after these two events trigger the pipeline will not trigger the pipeline again. Don't know if this is what we want

heevasti: This is how I see the work flow sh/could go:
1. A programmer wants to make changes and creates a new branch.
2. Changes are made in the branch and pushed until the programmer and pipeline are satisfied.
3. The programmer makes a pull request of the branch, which triggers to run the "full" pipeline for the first time
4. The programmer fixes possible problems coming from the full pipeline and sets the ticket for review
  4.1. Here the prgrmr has to test locally the issues from full pipeline are fixed. This is a possible caveat
5. The issue gets reviewed and programmer tackles the review comments.
  5.1. These changes could also affect the full pipeline. This is also a possible caveat. Local testing is needed.
6. The reviewer accepts the issue and the prgrmr sets it to (auto) merge. This runs the full pipeline again.
  6.1. If mistakes were made in the steps 4 or 5, they will come up here. The merge gets rejected in that case.
7. If the full pipeline caused issues, the progrmr needs to fix those and set it to (auto) merge again.
  7.1. I hope here no new approval by reviewer is needed...

heevasti (approved):
Great job Ravi!

PR #28, opened 2023-08-11
update_to_v041

heevasti:
Synchronizing with Gitlab for v0.41.0.


tests/tools/test_proc.py
```diff
@@ -41,7 +41,7 @@ def _start_qmi_context(self):
             "program_args": ["test_program_args"],
         }
         qmi.start("ContextName2", context_cfg={
-                "ContextName2": {"tcp_server_port": 511},
```
rbudhrani, line 44: this may be something we wish to note down in the CHANGELOG. The syncs from QMI are in the CHANGELOG, but this will be a diff. Maybe best to sync back to GitLab after this is merged

rbudhrani (dismissed):

rbudhrani (approved):

Issue #43, opened 2023-10-02
Implement `read_until_timeout` and `discard_read` in QMI_UsbTmcTransport class


heevasti:
Description 
=========== 
For the UsbTmc, the current `read` and/or `read_until` and/or `read_until_timeout` should (try to) return the whole `_read_buffer` AND whatever the `_read_message(timeout)` returns, up to their `nbytes` or `message_terminator` limits. This is not happening now. Also `discard_read` should get implemented.

1. The `read` method should be modified to work the same way as the `QMI_Transport.read` base class. It should also start with the existing buffer check
```python
        nbuf = len(self._read_buffer)
        if nbuf >= nbytes:
            # The requested number of bytes are already in the buffer.
            # Return them immediately.
            ret = bytes(self._read_buffer[:nbytes])
            self._read_buffer = self._read_buffer[nbytes:]
            return ret
```
And if no buffer is present, then check the timeout value as now and read-in the buffer.
```
        # USB requires a timeout
        if timeout is None:
            timeout = self.DEFAULT_READ_TIMEOUT

        # Read buffer was too short or is empty - read a new message from the instrument.
        self._read_buffer.extend(self._read_message(timeout))
```
Then re-check the buffer length. By too short buffer length discard buffer and raise exception. By too long data discard buffer after `nbytes` and return `nbytes` data like now. And the with `nbytes` of data return and clear whole data buffer.

2. The `read_until` (of the `QMI_UsbTmcTransport`) command says in the description that the `message_terminator` is ignored as it is not somehow valid for USBTMC protocol. This means that actually this command should not be implemented and be set with `NotImplementedError` and the `read_until_timeout` be implemented instead. The problem here is that instruments using SCPI protocol use `ScpiProtocol` class where in `ask` the `read_until` is used by default. We cannot change this to `read_until_timeout` there as also other transports can use this and that would mess things up. So, we would need to keep `read_until` but change the implementation such that it forwards then to `read_until_timeout`.

> **_NOTE:_** If there remains partial messages in the `self._read_buffer` and this keeps messing up the parsing of responses or similar, currently it is up to the user or e.g. implemented QMI driver commands to call `discard_read()` to reset the buffer.

3. Set the `discard_read` method of the `QMI_UsbTmcTransport` class to clear the current buffer (`self._read_buffer = bytes<array>()`).
```python
def discard_read():
    # We should empty the buffer, or if it is empty, discard the next message from the source
    if not len(self._read_buffer):
[...]

PR #45, opened 2023-10-03
43 implement read until timeout and discard read in qmi usbtmctransport class

heevasti:
Related to #43 

Acceptance criteria 
=================== 
- [x] Functionality implemented in agreement with description of the ticket
- [x] Tests are updated
- [x] Documentation updated

### DIFF
--- a/qmi/core/transport.py
+++ b/qmi/core/transport.py
@@ -854,6 +854,10 @@ class QMI_UsbTmcTransport(QMI_Transport):
                 is received before the requested number of bytes is reached.
         """
         self._check_is_open()
+        # USB requires a timeout
+        if timeout is None:
+            timeout = self.DEFAULT_READ_TIMEOUT
+
         nbuf = len(self._read_buffer)
         while True:
             if nbuf == nbytes:
@@ -861,10 +865,6 @@ class QMI_UsbTmcTransport(QMI_Transport):
                 ret = self._read_buffer
                 self._read_buffer = bytes()
                 return ret
-
-            # USB requires a timeout
-            if timeout is None:
-                timeout = self.DEFAULT_READ_TIMEOUT
 
             # Read buffer was of wrong length or is empty - read a new message from the instrument.
             self._read_buffer += self._read_message(timeout)
--- a/qmi/core/transport.py
+++ b/qmi/core/transport.py
@@ -803,7 +803,6 @@ class QMI_UsbTmcTransport(QMI_Transport):
         self.vendorid = vendorid
         self.productid = productid
         self.serialnr = serialnr
-        self._read_buffer = bytes()
 
     @staticmethod
     def _validate_product_id(productid):
@@ -858,23 +857,15 @@ class QMI_UsbTmcTransport(QMI_Transport):
         if timeout is None:
             timeout = self.DEFAULT_READ_TIMEOUT
 
-        nbuf = len(self._read_buffer)
-        while True:
-            if nbuf == nbytes:
-                # The requested number of bytes are already in the buffer. Return them immediately.
-                ret = self._read_buffer
-                self._read_buffer = bytes()
-                return ret
+        # Read a USB message
+        ret = self._read_message(timeout)
+        if len(ret) == nbytes:
+            return ret
 
-            # Read buffer was of wrong length or is empty - read a new message from the instrument.
-            self._read_buffer += self._read_message(timeout)
-            nbuf = len(self._read_buffer)
-            if nbuf != nbytes:
-                _logger.debug("USBTMC read buffer contained data %s" % self._read_buffer.decode())
-                self._read_buffer = bytes()
-                raise QMI_EndOfInputException(
-                    f"The read buffer did not contain expected bytes of data ({nbuf} != {nbytes}."
-                )
+        _logger.debug("USBTMC read message contained data %s" % ret.decode())
+        raise QMI_EndOfInputException(
+            f"The read message did not contain expected bytes of data ({len(ret)} != {nbytes}."
+        )
 
     def read_until(self, message_terminator: bytes, timeout: Optional[float]) -> bytes:
         """The specified message_terminator is ignored. Instead, the instrument must autonomously indicate the end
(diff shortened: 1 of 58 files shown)


qmi/core/transport.py
```diff
@@ -858,70 +853,105 @@ def read(self, nbytes: int, timeout: Optional[float]) -> bytes:
...
 
-        if not self._read_buffer:
-            # Read buffer is empty - read a new message from the instrument.
-            self._read_buffer = self._read_message(timeout)
-
-        # Check that the buffer contains enough data.
-        if len(self._read_buffer) < nbytes:
-            raise QMI_EndOfInputException("Expecting {} bytes but only {} bytes remaining in USBTMC message"
-                                          .format(nbytes, len(self._read_buffer)))
+        nbuf = len(self._read_buffer)
```
pieterjbotma, line 861: I don't understand what the difference is between the removed and added code? Both seem to read message until timeout, and throw error if length is not what is expected?

qmi/core/transport.py
```diff
@@ -858,70 +853,105 @@ def read(self, nbytes: int, timeout: Optional[float]) -> bytes:
...
-            # Read buffer is empty - read a new message from the instrument.
-            data = self._read_message(timeout)
+            # Read a new message from the instrument.
+            data += self._read_message(timeout)
 
         return data
 
+    def discard_read(self) -> None:
+        # We should empty the buffer, or if it is empty, discard the next message from the source
+        if not len(self._read_buffer):
```
pieterjbotma, line 934: Why choose? Shouldn't you clear the 'read_buffer' *and* clear the message buffer?

qmi/core/transport.py
```diff
@@ -858,70 +853,105 @@ def read(self, nbytes: int, timeout: Optional[float]) -> bytes:
...
+            # Read buffer is empty - read a new message from the instrument.
+            try:
+                self._read_message(0.0)
```
pieterjbotma, line 937: what happens if there's multiple message in the queue? don't you want to read until timeout?

qmi/core/transport.py
```diff
@@ -858,70 +853,105 @@ def read(self, nbytes: int, timeout: Optional[float]) -> bytes:
...
 
         # USB requires a timeout
         if timeout is None:
             timeout = self.DEFAULT_READ_TIMEOUT
 
+        data = bytes()
         if self._read_buffer:
             # Data already available in buffer; return it now.
-            data = self._read_buffer
+            data += self._read_buffer
```
pieterjbotma, line 923: really minor: but what was reason for adding the `data` and `+=`?

pieterjbotma (dismissed):

qmi/core/transport.py
```diff
@@ -858,70 +853,105 @@ def read(self, nbytes: int, timeout: Optional[float]) -> bytes:
...
+
+        If there is already data in the buffer, return it.
 
         If the timeout expires before the message is received, the read is
         aborted and any data already received are discarded. In this
         case QMI_TimeoutException is raised.
 
         Parameters:
-            message_terminator: Message terminator characters.
+            nbytes: This input is ignored.
```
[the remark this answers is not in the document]
heevasti: Mainly to fulfil compatibility with parent `QMI_Transport` class. It has the abstract function for this. Also we don't want to break the API yet, it'll be done in the other ticket with name changes.

qmi/core/transport.py
heevasti: Yes, you're right, it is not needed as we take over a (byte) string and not something mutable like a list. So doing `self._read_buffer = bytes()` does not clear the overtaken string (`data = self._read_buffer`) by reference. I will change it back.

qmi/core/transport.py
heevasti: USB should not have multiple messages in the queue, but anyhow if the user wants to discard multiple reads, then is should be called as many times instead of clearing everything. I think this is more the idea of this function.

qmi/core/transport.py
pieterjbotma: but this part of the code will throw an exception if the first read doesn't return the correct amount of bytes:
```python
    if nbuf != nbytes:
        _logger.debug("USBTMC read buffer contained data %s" % self._read_buffer.decode())
        self._read_buffer = bytes()
        raise QMI_EndOfInputException(
            f"The read buffer did not contain expected bytes of data ({nbuf} != {nbytes}."
        )
```

heevasti: That is correct. The `while True:` indeed is quite useless. I will remove it.

What comes to the other changes is that I changed the order to return the data from the buffer directly if there is exactly the correct amount of bytes in. Any wrong number of data, or no data, should result in reading a new message. What I realize here now is that 
a) The read buffer should be at that point cleared as the USB should only return whole messages and it should not be possible(?) that only partial messages are returned unless TimeoutError is raised. 
b) the new message read in should not be added to existing read buffer, but checked directly if it has the expected amount of  bytes.

I think I had the original `while` loop idea as being for cases where less than expected `nbytes` data has been delivered by (multiple) read(s) and it would loop and add the bytes into the read buffer until at least `nbytes` are available, unless it gets some exception. But the function can only do a partial readout only in case of a timeout (as USB should always return one whole message), which means that an exception is raised and the `while` loop will be broken - so the re-read and adding to the `_read_buffer` cannot work. But I then somehow forgot the whole thing and didn't remove it.

heevasti: Perhaps saving the read into the `_read_buffer`, and then calling to read again could be a solution, but the correct way would rather be to increase the timeout such that the whole message will be read in and not only partially.

heevasti: If going for the solution of increasing timeout, we can actually remove the whole `self._read_buffer` as this function was the only one actually adding data to it. If require rather that the user uses longer timeout next time instead of allowing it to place several read calls in sequence, we can remove it. Do you agree?

qmi/core/transport.py
heevasti: In the end I would suggest to remove this, w.r.t. comment above about removing `self._read_buffer` for the USBTMC transport altogether.

qmi/core/transport.py
pieterjbotma: Yep. let's go with removing the `_read_buffer`.  That simplifies things a bit and keeps it closer to how usbtmc works in reality. Then the user has multiple options. Either read a message it knows how long it will be, or just read a random message with an appropriate timeout.

This does make `discard_read` pointless again? So we can either have that just be pass, or raise not implemented?

heevasti: `discard_read` can still try to read from the instrument and discard if it got something as a reply.

rbudhrani (approved):

Issue #44, opened 2023-10-02
Change implementation for `read` and `read_until[_timeout]` methods for `QMI_Vxi11Transport` class


heevasti:
Description 
=========== 
The implementation of the Vxi11 protocol in the `QMI_Vxi11Transport` class has a couple of shortcomings. The `read` method tries to read `nbytes` data, but if the protocol does not return enough data, we get `Vxi11Exception`. The read data gets in that case discarded. For `read_until` method, the situation is quite the same, with the addition that by successful read, but the last character being some other than the expected `message_terminator`, exception is raised and data discarded. There is no implementation of the `read_until_timeout`.

Meanwhile in `discard_read` method data is read one byte at a time from the device, and then eventually any data read is discarded. 

We could rather make use of the similar scheme as for other "Transports", where data is read into `self._read_buffer`. We can do this with the approach of `discard_read` but now adding each read character into the `self._read_buffer`. For `read` case until `nbytes` bytes are in the buffer, and for `read_until` until `message_terminator` is reached:
```python
    def read(...):
        nbuf = len(self._read_buffer)
        if nbuf >= nbytes:
            # The requested number of bytes are already in the buffer.
            # Return them immediately.
            ret = bytes(self._read_buffer[:nbytes])
            self._read_buffer = self._read_buffer[nbytes:]
            return ret

        # timeout handling here, and then
        while len(self._read_buffer) < nbytes:
            try:
                self._read_buffer.extend(self._safe_serial.read_raw(1))
            ...

        return self._read_buffer
```
```python
    def read_until(...):
        nbuf = len(self._read_buffer)
        if nbuf > 0 and message_terminator in self._read_buffer:
            # The requested response is already in the buffer.
            # Return it immediately.
            terminator_index = self._read_buffer.index(message_terminator)
            ret = bytes(self._read_buffer[:terminator_index])
            self._read_buffer = self._read_buffer[terminator_index:]
            return ret

        try:
            while True:
                self._read_buffer.extend(self._safe_serial.read_raw())
                # Validate terminator.
                data_term_char = self._read_buffer[-1:]  # use slice rather than index to get back bytes()
                if data_term_char == message_terminator:
                    break

        # terminator and timeout handling here as before, and then

        ret = self._read_buffer
        self._read_buffer = bytearray()
        return ret
```
and finally we could add implementation
```python
    def read_until_timeout(...):
        return read(nbytes, timeout)
```
[...]

PR #46, opened 2023-10-17
QMI-044 change implementation for read and read until timeout methods for qmi vxi11transport class

heevasti:
Related to #44 

Acceptance criteria 
=================== 
- [ ] Functionality implemented in agreement with description of the ticket
- [ ] Tests are updated
- [ ] Documentation updated

### DIFF
--- a/qmi/core/transport.py
+++ b/qmi/core/transport.py
@@ -963,6 +963,7 @@ class QMI_Vxi11Transport(QMI_Transport):
 
         self._host = host
         self._instr: Optional[vxi11.Instrument] = None
+        self._read_buffer = bytes()
 
     @property
     def _safe_instr(self) -> vxi11.Instrument:
@@ -1007,21 +1008,34 @@ class QMI_Vxi11Transport(QMI_Transport):
 
     def read(self, nbytes: int, timeout: Optional[float]) -> bytes:
         self._check_is_open()
+        nbuf = len(self._read_buffer)
+        if nbuf >= nbytes:
+            # The requested number of bytes are already in the buffer. Return them immediately.
+            ret = self._read_buffer[:nbytes]
+            self._read_buffer = self._read_buffer[nbytes:]
+            return ret
+
         old_timeout = self._safe_instr.timeout
         if timeout:
             self._safe_instr.timeout = timeout
+
         try:
-            data = self._safe_instr.read_raw(nbytes)
+            while len(self._read_buffer) < nbytes:
+                self._read_buffer += self._safe_instr.read_raw(1)
+
         except vxi11.vxi11.Vxi11Exception as err:
             if err.err == 15:
                 # Raise timeout exception separately to provide opportunity for upper layer to handle it.
-                raise QMI_TimeoutException("Timeout error attempting to read {} bytes".format(nbytes)) from err
+                raise QMI_TimeoutException(f"Timeout error attempting to read {nbytes} bytes") from err
             else:
-                raise QMI_InstrumentException("Error attempting to read {} bytes.".format(nbytes)) from err
+                raise QMI_EndOfInputException(f"Error attempting to read {nbytes} bytes.") from err
+
         finally:
             self._safe_instr.timeout = old_timeout
 
-        return data
+        ret = self._read_buffer
+        self._read_buffer = bytes()
+        return ret
 
     def read_until(self, message_terminator: bytes, timeout: Optional[float]) -> bytes:
         self._check_is_open()
@@ -1029,6 +1043,14 @@ class QMI_Vxi11Transport(QMI_Transport):
             raise QMI_InstrumentException(
                 "VXI11 instrument only support 1 byte terminating character, received {!r}.".format(message_terminator)
             )
+
+        nbuf = len(self._read_buffer)
+        if nbuf > 0 and message_terminator in self._read_buffer:
+            # The requested response is already in the buffer. Return it immediately.
+            terminator_index = self._read_buffer.index(message_terminator)
+            ret = self._read_buffer[:terminator_index]
+            self._read_buffer = self._read_buffer[terminator_index:]
+            return ret
 
         # Set terminator, but keep old value.
         old_term_char = self._safe_instr.term_char
@@ -1039,13 +1061,11 @@ class QMI_Vxi11Transport(QMI_Transport):
         if timeout:
             self._safe_instr.timeout = timeout
 
-        # Attempt to read the data.
         try:
-            data = b''
             while True:
-                data += self._safe_instr.read_raw()
+                self._read_buffer += self._safe_instr.read_raw(1)
                 # Validate terminator.
-                data_term_char = data[-1:]  # use slice rather than index to get back bytes()
+                data_term_char = self._read_buffer[-1:]  # use slice rather than index to get back bytes()
                 if data_term_char == message_terminator:
                     break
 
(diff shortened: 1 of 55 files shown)


qmi/core/transport.py
```diff
@@ -1071,29 +1091,40 @@ def read_until(self, message_terminator: bytes, timeout: Optional[float]) -> byt
...
             self._safe_instr.term_char = old_term_char
             self._safe_instr.timeout = old_timeout
 
-        return data
+        ret = self._read_buffer
+        self._read_buffer = bytes()
+        return ret
+
+    def read_until_timeout(self, nbytes: int, timeout: float):
+        return self.read(nbytes, timeout)
```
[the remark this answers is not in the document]
heevasti: Right, QMI_TimeoutException should be caught indeed. Then it can return whatever was read in the buffer until that exception and clear the buffer. In any other exception, if something was read in, it'll stay in the `_read_buffer`.

qmi/core/transport.py
```diff
@@ -1019,31 +1020,52 @@ def write(self, data: bytes) -> None:
         except vxi11.vxi11.Vxi11Exception as err:
             raise QMI_InstrumentException() from err
 
-    def read(self, nbytes: int, timeout: Optional[float]) -> bytes:
+    def read(self, nbytes: int, timeout: Optional[float] = None) -> bytes:
         self._check_is_open()
+        nbuf = len(self._read_buffer)
+        if nbuf >= nbytes:
```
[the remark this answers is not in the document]
heevasti: Well the idea I had for this ticket was exactly that. If you get only partial read of some (big) response, you can read again until all the response has been read. This will require some check on the receiving script to see how much data was read in and how much should still be read. There should in this case be no discarding of data be done in between. This should also prevent the issues when not all data was read from the instrument, but the next query is nevertheless sent and it starts with the remainder data of the previous query/read, and the response is thus messed up.

The point is, there was no option to do it like this before. With this, this possibility is enabled. To utilize it, it does need the scripts to have the intelligence to handle partial readout returns (like a `while` loop or some `if` checks), before continuing to the next query. Currently, the only thing you can do is to discard the already read partial data with `discard_read` and then try again with a longer timeout. With interactive mode script that is still quite OK, but with static scripts it probably means you have to change the timeout conf/setting and then run the script again.

It might make the existing scripts less resilient, but with this change issues with asynchronous responses (like we've seen sometimes on Tektronix) on those scripts can be more effectively tackled. They just then need to be adjusted to use this new feature. Too bad this is not possible with the USB devices as they send the whole message or nothing.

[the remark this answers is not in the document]
pieterjbotma: I'm a bit afraid of the performance implications here. Calling back to python for every byte is going to be quite slow. Could you maybe test this first on a device to see the impact on a big read.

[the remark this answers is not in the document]
pieterjbotma: Maybe we can have the normal `read` just read the full amount of bytes for maximum performance. Then for `read_until` and `read_untill_timeout` we can have it read byte for byte?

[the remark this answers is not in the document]
heevasti: You're right, reading byte-per-byte turned out to be very slow. In my tests, reading 156536 bytes (from several separate calls using the `read_until` method for an instrument) took ~11s when using the default `read_raw()` call. Meanwhile, reading the same amount of bytes with `read_raw(1)` took 75 seconds! But, when using `read_raw(512)`, the reading took also just over 11 seconds. So we could go for batches of 512 bytes instead, which does help to avoid timeout errors as each batch read has its own timeout count, instead of a singular call for unknown number of bytes until timeout or `message_terminator`. This was tested with `read_until`, a similar test needs to be done for `read` method where we read up to `nbytes` bytes with batches of `min(512, nbytes - bytes_read)`. To be commented further after the test.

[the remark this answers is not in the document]
heevasti: A test on `read(nbytes)` method reading **`read_raw(nbytes)`**, with total of 155548 bytes divided in several calls, took about  **0.26 seconds**. The original version of the test with `read_until` took 3.8 seconds and with batches of 512 bytes it took 4.0 seconds, as comparison.

Batched **`read(min(512, nbytes - bytes_read))`** took **0.72 seconds**

byte-per-byte read took whopping 46 seconds!

So clearly the byte-per-byte read is not doable with even relatively small data. read_until is also much slower (up to 20x) than read(nbytes), but as we really cannot know beforehand in many cases how many bytes we will read, we'll have to use read_until. But another point is that the batch read with while True: read(min(512, nbytes-len(self._read_buffer))) was not so much slower than (2.5x...3x) direct read. With 60s timeout with direct read we can read in max 36.5MB of data, while there is no such limitation with the batch read. If we want to read 36.6MB data we can, although it'll take almost 3 minutes.

qmi/core/transport.py
```diff
@@ -1019,31 +1020,53 @@ def write(self, data: bytes) -> None:
...
+            return ret
+
         old_timeout = self._safe_instr.timeout
         if timeout:
             self._safe_instr.timeout = timeout
+
         try:
-            data = self._safe_instr.read_raw(nbytes)
+            while len(self._read_buffer) < nbytes:
+                nbytes_batch = min(512, nbytes - len(self._read_buffer))
```
pieterjbotma, line 1038: minor: can't we replace 512 with `self._safe_instr.max_recv_size`. Then we stick to what the driver can handle?

qmi/core/transport.py
```diff
@@ -1071,29 +1092,47 @@ def read_until(self, message_terminator: bytes, timeout: Optional[float]) -> byt
...
             except vxi11.vxi11.Vxi11Exception as err:
                 if err.err == 15:
                     break
                 else:
+                    self._safe_instr.timeout = old_timeout
                     raise QMI_InstrumentException("Error attempting to read a byte") from err
+
             except TimeoutError:
                 break
-            finally:
```
pieterjbotma, line 1095: minor: why did you remove this finally? seems like a better way to make sure the old timeout is restored?

qmi/core/transport.py
```diff
@@ -1071,29 +1092,47 @@ def read_until(self, message_terminator: bytes, timeout: Optional[float]) -> byt
...
         self._check_is_open()
         old_timeout = self._safe_instr.timeout
-        self._safe_instr.timeout = 0.0
+        self._safe_instr.timeout = 0.0  # Immediate read
+        # Clear any possible data in read buffer first
+        self._read_buffer = bytes()
         while True:
             try:
-                _ = self._safe_instr.read_raw(1)
+                self._safe_instr.read_raw(1)
```
pieterjbotma, line 1123: wouldn't it also be faster to just do `self._safe_instr.read_raw()` to read the full chunk size?

qmi/core/transport.py
```diff
@@ -1019,31 +1020,53 @@ def write(self, data: bytes) -> None:
...
+            self._read_buffer = self._read_buffer[nbytes:]
+            return ret
+
         old_timeout = self._safe_instr.timeout
         if timeout:
             self._safe_instr.timeout = timeout
+
         try:
-            data = self._safe_instr.read_raw(nbytes)
+            while len(self._read_buffer) < nbytes:
```
pieterjbotma, line 1037: minor/major: only problem is that we're not really respecting our timeouts. if the batch is read faster than the timeout, we can basically read indefinitely.

pieterjbotma (changes requested):

qmi/core/transport.py
heevasti: Yes, that sounds like a good idea!

qmi/core/transport.py
heevasti: ... up to `nbytes` indefinitely. So, not really an issue as we have a max size limit.

qmi/core/transport.py
heevasti: It would be faster, but I'm not sure what was the original idea behind reading this per byte. Perhaps to make sure everything gets discarded, even if some message terminator is in the string or such?

qmi/core/transport.py
heevasti: Because the try-except is in a `while` loop and doesn't each `try` finish with `finally`, resetting the timeout after the first  loop?

qmi/core/transport.py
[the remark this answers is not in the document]
heevasti: Changed to `self._instr.max_recv_size` as suggested in another comment.

pieterjbotma (dismissed):

pieterjbotma (approved):

PR #49, opened 2023-10-23
Fix for readthedocs

QFer:
* comply to v2 of readthedocs.yaml


heevasti (approved):
Thanks for fixing this. I guess we will upgrade also the Python version soonish, when the support for 3.8 is finished.

Issue #58, opened 2023-12-18
Create UDP transport and class


heevasti:
Description 
=========== 
There is a new request to have UDP transport class for Tenma power supply instruments. As UDP is quite similar to TCP, basically TCP with less checks and stability, we should be able to quickly create one based on the TCP class with stripping some details. Then, we should make a **base class** that works for both the UDP and TCP transport classes, so to save some double code.

First thing in `qmi.core.transport` module is the transport descriptor parses definition. This probably can be exactly the same as for TCP, with changing the the names only:
```python
UdpTransportDescriptorParser = TransportDescriptorParser(
    "udp",
    [("host", (str, True)),
     ('port', (int, True))],
    {'connect_timeout': (float, False)}
)
```

<p>
<details>
<summary>Cees Wolfs has already tested the UDP communication with the following transport class:</summary>
  
```python
class QMI_UdpTransport(QMI_Transport):
    """Bidirectional byte stream via UDP network connection.

    An instance of QMI_UdpTransport represents a client-side UDP connection
    to an instrument. Server-side UDP connections are not supported.
    """

    DEFAULT_CONNECT_TIMEOUT = 10

    def __init__(self, host: str, port: int, connect_timeout: Optional[float] = DEFAULT_CONNECT_TIMEOUT) -> None:
        """Initialize the UDP transport by setting up the connection details.

        Parameters:
            connect_timeout: Maximum time to connect in seconds.

        Raises:
            ~qmi.core.exceptions.QMI_TimeoutException: If connecting takes longer than the specified connection
            timeout.
        """
        super().__init__()

        self._validate_host(host)
        self._validate_udp_port(port)

        self._address = (host, port)
        self._connect_timeout = connect_timeout
        self._socket: Optional[socket.socket] = None
        self._read_buffer = bytearray()

    @staticmethod
    def _validate_udp_port(port):
        if port < 1 or port > 65535:
            raise QMI_TransportDescriptorException("Invalid UDP port number {}".format(port))

    @staticmethod
    def _validate_host(host):
        if (not _is_valid_hostname(host)) and (not _is_valid_ipaddress(host)):
            raise QMI_TransportDescriptorException("Invalid host name {}".format(host))

    def __str__(self) -> str:
        remote_addr = format_address_and_port(self._address)
        return "QMI_UdpTransport(remote={})".format(remote_addr)

    @property
    def _safe_socket(self) -> socket.socket:
[...]

PR #59, opened 2023-12-18
Create udp transport and class

heevasti:
Related to #58

Acceptance criteria 
=================== 
- [ ] Functionality implemented in agreement with description of the ticket
- [ ] Tests are updated
- [ ] Documentation updated

### DIFF
--- a/qmi/core/transport.py
+++ b/qmi/core/transport.py
@@ -691,6 +691,7 @@ class QMI_UdpTcpTransportBase(QMI_Transport):
                     nbuf = len(self._read_buffer)
                     raise QMI_TimeoutException(f"Function timeout after {nbuf} bytes without message terminator")
                 self._safe_socket.settimeout(tremain)
+
             # Read from socket.
             try:
                 # NOTE: Reading up to 512 bytes here for TCP, as larger packet size
@@ -701,12 +702,13 @@ class QMI_UdpTcpTransportBase(QMI_Transport):
                     b, addr = self._safe_socket.recvfrom(4096)
                 else:
                     b, addr = self._safe_socket.recvfrom(512)
-            except (BlockingIOError, socket.timeout):
+            except (BlockingIOError, socket.timeout, TimeoutError):
                 # timeout in socket.recv()
                 nbuf = len(self._read_buffer)
                 raise QMI_TimeoutException(f"Socket timeout after {nbuf} bytes without message terminator")
             except OSError:
                 raise QMI_RuntimeException(f"UDP packet size was larger than 4096. Data is lost.")
+
             if self._assert_addr and addr != self._address:
                 _logger.warning(f"Received data from address %s while expected data only from %s!", addr, self._address)
                 del b
--- a/qmi/core/transport.py
+++ b/qmi/core/transport.py
@@ -787,8 +787,8 @@ class QMI_UdpTransport(QMI_UdpTcpTransportBase):
         socket.gethostbyname(self._address[0])
         # Create socket
         self._socket = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
-        # To set the local address to point to our client, we need to bind it. To avoid OSError, we use port number + 1
-        self._socket.bind((self._address[0], self._address[1] + 1))
+        # To set the local address to point to our client, we need to bind it.
+        self._socket.bind(("", self._address[1]))
 
     def close(self) -> None:
         _logger.debug("Closing UDP transport %s", self)
--- a/qmi/core/transport.py
+++ b/qmi/core/transport.py
@@ -13,7 +13,7 @@ import serial
 from qmi.core.context import QMI_Context
 from qmi.core.exceptions import (
     QMI_InvalidOperationException, QMI_TimeoutException, QMI_EndOfInputException,
-    QMI_TransportDescriptorException, QMI_InstrumentException
+    QMI_TransportDescriptorException, QMI_InstrumentException, QMI_RuntimeException
 )
 from qmi.core.util import format_address_and_port
 
@@ -579,7 +579,7 @@ class QMI_UdpTcpTransportBase(QMI_Transport):
         self._validate_host(host)
         self._validate_port(port)
         self._address = (host, port)
-        self._assert_addr = True if assert_address else False
+        self._assert_addr = True if assert_address else False  # For UDP, as it accepts data from any client.
         self._socket: Optional[socket.socket] = None
         self._read_buffer = bytearray()
 
@@ -630,13 +630,28 @@ class QMI_UdpTcpTransportBase(QMI_Transport):
         while nbuf < nbytes:
             self._safe_socket.settimeout(tremain)
             try:
-                b, addr = self._safe_socket.recvfrom(nbytes - nbuf)
+                if self._assert_addr:
+                    # For UDP, reading data fails if the arriving packet size is larger than the given nbytes (- nbuf)
+                    # to read. To avoid this happening, read at least 4096 standard that is larger than regular MTU.
+                    udp_packet_size = max(nbytes - nbuf, 4096)
+                    try:
+                        b, addr = self._safe_socket.recvfrom(udp_packet_size)
+                    except (BlockingIOError, socket.timeout) as err:
+                        raise err  # Re-raise this for handling in the upper try-except.
+                    except OSError:
+                        raise QMI_RuntimeException(f"UDP packet size was larger than {udp_packet_size}. Data is lost.")
+
+                    if addr != self._address:
+                        _logger.warning(f"Received data from address %s while expected data only from %s!", addr,
+                                        self._address)
+                        del b
+                        continue
+
+                else:
+                    b, addr = self._safe_socket.recvfrom(nbytes - nbuf)
+
             except (BlockingIOError, socket.timeout):
                 raise QMI_TimeoutException(f"Timeout after {nbuf} bytes while expecting {nbytes}")
-            if self._assert_addr and addr != self._address:
-                _logger.warning(f"Received data from address %s while expected data only from %s!", addr, self._address)
-                del b
-                continue
             if not b:
                 raise QMI_EndOfInputException(
                     f"Reached end of input from socket {format_address_and_port(self._address)}")
@@ -678,14 +693,20 @@ class QMI_UdpTcpTransportBase(QMI_Transport):
                 self._safe_socket.settimeout(tremain)
             # Read from socket.
             try:
-                # NOTE: Reading up to 4096 bytes here.
+                # NOTE: Reading up to 512 bytes here for TCP, as larger packet size
+                #       causes issues with Tektronix 5014 AWG. For UDP read 4096.
                 #       Exact amount does not matter but a small power of 2 is recommended, e.g. 4096
                 #       by socket.recv() documentation.
-                b, addr = self._safe_socket.recvfrom(512)
+                if self._assert_addr:
+                    b, addr = self._safe_socket.recvfrom(4096)
+                else:
+                    b, addr = self._safe_socket.recvfrom(512)
             except (BlockingIOError, socket.timeout):
                 # timeout in socket.recv()
                 nbuf = len(self._read_buffer)
                 raise QMI_TimeoutException(f"Socket timeout after {nbuf} bytes without message terminator")
+            except OSError:
+                raise QMI_RuntimeException(f"UDP packet size was larger than 4096. Data is lost.")
             if self._assert_addr and addr != self._address:
                 _logger.warning(f"Received data from address %s while expected data only from %s!", addr, self._address)
                 del b
@@ -764,8 +785,12 @@ class QMI_UdpTransport(QMI_UdpTcpTransportBase):
 
     def _open_transport(self) -> None:
         super()._open_transport()
+        # Validate the host exists before trying to create a socket
+        socket.gethostbyname(self._address[0])
         # Create socket
         self._socket = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+        # To set the local address to point to our client, we need to bind it. To avoid OSError, we use port number + 1
+        self._socket.bind((self._address[0], self._address[1] + 1))
 
     def close(self) -> None:
         _logger.debug("Closing UDP transport %s", self)
@@ -776,54 +801,6 @@ class QMI_UdpTransport(QMI_UdpTcpTransportBase):
         # NOTE: We explicitly adjust the socket timeout before each send/recv call.
         self._safe_socket.settimeout(None)
         self._safe_socket.sendto(data, self._address)
-
-    def read_until(self, message_terminator: bytes, timeout: Optional[float]) -> bytes:
-        self._check_is_open()
-        tstart = time.monotonic()
-        first_pass = True
-        print("self._address", self._address)
-        print("self._safe_socket.getsockname", self._safe_socket.getsockname())
-        while True:
-            # Check if message already received.
-            p = self._read_buffer.find(message_terminator)
-            if p >= 0:
-                # Found "message_terminator" - return data.
-                nbytes = p + len(message_terminator)
-                ret = bytes(self._read_buffer[:nbytes])
-                self._read_buffer = self._read_buffer[nbytes:]
-                return ret
-            # Determine timeout.
-            if first_pass:
-                # Try to read at least once, even if timeout is zero.
-                self._safe_socket.settimeout(timeout)
-                first_pass = False
-            elif timeout is not None:
-                # Calculate remaining time.
-                tremain = tstart + timeout - time.monotonic()
-                if tremain < 0:
-                    nbuf = len(self._read_buffer)
-                    raise QMI_TimeoutException(f"Function timeout after {nbuf} bytes without message terminator")
-                self._safe_socket.settimeout(tremain)
-            # Read from socket.
-            try:
-                # NOTE: Reading up to 4096 bytes here.
-                #       Exact amount does not matter but a small power of 2 is recommended, e.g. 4096
-                #       by socket.recv() documentation.
-                b, addr = self._safe_socket.recvfrom(512)
-                print("addr", addr)
-            except (BlockingIOError, socket.timeout):
-                # timeout in socket.recv()
-                nbuf = len(self._read_buffer)
-                raise QMI_TimeoutException(f"Socket timeout after {nbuf} bytes without message terminator")
-            if self._assert_addr and addr != self._address:
-                _logger.warning(f"Received data from address %s while expected data only from %s!", addr, self._address)
-                del b
-                continue
-            if not b:
-                # end of stream
-                raise QMI_EndOfInputException(
-                    f"Reached end of input from socket {format_address_and_port(self._address)}")
-            self._read_buffer.extend(b)
 
 
 class QMI_TcpTransport(QMI_UdpTcpTransportBase):
--- a/qmi/core/transport.py
+++ b/qmi/core/transport.py
@@ -588,7 +588,7 @@ def _is_valid_ipaddress(address: str) -> bool:
     return False
 
 
-class QMI_UdpTcpTransportBase(QMI_Transport):
+class QMI_SocketTransport(QMI_Transport):
     """Base class for bidirectional data streams via UDP or TCP network connection."""
 
     def __init__(self, host: str, port: int) -> None:
@@ -715,7 +715,7 @@ class QMI_UdpTcpTransportBase(QMI_Transport):
                 break
 
 
-class QMI_UdpTransport(QMI_UdpTcpTransportBase):
+class QMI_UdpTransport(QMI_SocketTransport):
     """
     An instance of QMI_UdpTransport represents a server-side UDP connection with
     a listening port to an instrument. Client-side UDP connections are not supported.
@@ -734,11 +734,10 @@ class QMI_UdpTransport(QMI_UdpTcpTransportBase):
             host: The server IP address.
             port: The port for the address.
         """
-        self._validate_udp_port(port)
         super().__init__(host, port)
 
-    @staticmethod
-    def _validate_udp_port(port):
+    def _validate_port(self, port):
+        super()._validate_port(port)
         if port == QMI_Context.DEFAULT_UDP_RESPONDER_PORT:
             raise QMI_TransportDescriptorException(f"UDP port number {port} not allowed")
 
@@ -829,7 +828,7 @@ class QMI_UdpTransport(QMI_UdpTcpTransportBase):
                 self._safe_socket.settimeout(tremain)
 
 
-class QMI_TcpTransport(QMI_UdpTcpTransportBase):
+class QMI_TcpTransport(QMI_SocketTransport):
     """Bidirectional byte stream via TCP network connection.
 
     An instance of QMI_TcpTransport represents a client-side TCP connection
--- a/qmi/core/transport.py
+++ b/qmi/core/transport.py
@@ -655,6 +655,8 @@ class QMI_UdpTcpTransportBase(QMI_Transport):
             if not b:
                 raise QMI_EndOfInputException(
                     f"Reached end of input from socket {format_address_and_port(self._address)}")
+            if len(b) > 99:
+                print(addr, b)
             self._read_buffer.extend(b)
             nbuf = len(self._read_buffer)
             if timeout is not None:
--- a/qmi/core/transport.py
+++ b/qmi/core/transport.py
@@ -568,8 +568,8 @@ class QMI_UdpTcpTransportBase(QMI_Transport):
         """Initialize the UDP or TCP transport by connecting to the specified address.
 
         Parameters:
-            host: The server/client IP address.
-            port: The port for the address.
+            host:           The server/client IP address.
+            port:           The port for the address.
             assert_address: Boolean to indicate if the address where data was received from should be asserted.
         """
         super().__init__()
@@ -579,7 +579,7 @@ class QMI_UdpTcpTransportBase(QMI_Transport):
         self._validate_host(host)
         self._validate_port(port)
         self._address = (host, port)
-        self._assert_addr = True if assert_address else False  # For UDP, as it accepts data from any client.
+        self._assert_addr = assert_address  # For UDP, as it accepts data from any client.
         self._socket: Optional[socket.socket] = None
         self._read_buffer = bytearray()
 
@@ -638,8 +638,10 @@ class QMI_UdpTcpTransportBase(QMI_Transport):
                         b, addr = self._safe_socket.recvfrom(udp_packet_size)
                     except (BlockingIOError, socket.timeout) as err:
                         raise err  # Re-raise this for handling in the upper try-except.
-                    except OSError:
-                        raise QMI_RuntimeException(f"UDP packet size was larger than {udp_packet_size}. Data is lost.")
+                    except OSError as err:
+                        raise QMI_RuntimeException(
+                            f"UDP packet size was larger than {udp_packet_size}. Data is lost."
+                        ) from err
 
                     if addr != self._address:
                         _logger.warning("Received data from address %s while expected data only from %s!", addr,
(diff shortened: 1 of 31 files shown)


qmi/core/transport.py
```diff
@@ -647,18 +630,38 @@ def read(self, nbytes: int, timeout: Optional[float]) -> bytes:
...
-                b = self._safe_socket.recv(nbytes - nbuf)
+                if self._assert_addr:
+                    # For UDP, reading data fails if the arriving packet size is larger than the given nbytes (- nbuf)
+                    # to read. To avoid this happening, read at least 4096 standard that is larger than regular MTU.
+                    udp_packet_size = max(nbytes - nbuf, 4096)
+                    try:
+                        b, addr = self._safe_socket.recvfrom(udp_packet_size)
+                    except (BlockingIOError, socket.timeout) as err:
+                        raise err  # Re-raise this for handling in the upper try-except.
+                    except OSError:
```
rbudhrani, line 641: just like the previous exception, it would be good to re raise this one as `except OSError as exc` and then `raise QMI_RuntimeException(...) from exc`

qmi/core/transport.py
```diff
@@ -647,18 +630,38 @@ def read(self, nbytes: int, timeout: Optional[float]) -> bytes:
...
+                        raise QMI_RuntimeException(f"UDP packet size was larger than {udp_packet_size}. Data is lost.")
+
+                    if addr != self._address:
+                        _logger.warning(f"Received data from address %s while expected data only from %s!", addr,
```
rbudhrani, line 645: you are using an f-string but are providing % operator. Since we are logging we can use lazy formatting. This should work:
`_logger.warning("Received data from address %s while expected data only from %s!", addr,
                                        self._address)`

qmi/core/transport.py
```diff
@@ -647,18 +630,38 @@ def read(self, nbytes: int, timeout: Optional[float]) -> bytes:
...
+                                        self._address)
+                        del b
+                        continue
+
+                else:
+                    b, addr = self._safe_socket.recvfrom(nbytes - nbuf)
+
             except (BlockingIOError, socket.timeout):
```
rbudhrani, line 653: Let's explicitly re-raise here as well

qmi/core/transport.py
```diff
@@ -686,22 +689,34 @@ def read_until(self, message_terminator: bytes, timeout: Optional[float]) -> byt
...
                 if tremain < 0:
                     nbuf = len(self._read_buffer)
-                    raise QMI_TimeoutException("Timeout after {} bytes without message terminator".format(nbuf))
+                    raise QMI_TimeoutException(f"Function timeout after {nbuf} bytes without message terminator")
                 self._safe_socket.settimeout(tremain)
+
             # Read from socket.
             try:
-                # NOTE: Reading up to 4096 bytes here.
+                # NOTE: Reading up to 512 bytes here for TCP, as larger packet size
```
rbudhrani, line 697: should we really have this limit for only one instrument? Can we have an option in `read_until` to specify how many bytes we read?

qmi/core/transport.py
```diff
@@ -686,22 +689,34 @@ def read_until(self, message_terminator: bytes, timeout: Optional[float]) -> byt
...
+                #       causes issues with Tektronix 5014 AWG. For UDP read 4096.
                 #       Exact amount does not matter but a small power of 2 is recommended, e.g. 4096
                 #       by socket.recv() documentation.
-                b = self._safe_socket.recv(512)
-            except (BlockingIOError, socket.timeout):
+                if self._assert_addr:
+                    b, addr = self._safe_socket.recvfrom(4096)
+                else:
+                    b, addr = self._safe_socket.recvfrom(512)
+            except (BlockingIOError, socket.timeout, TimeoutError):
```
rbudhrani, line 705: let's explicitly re-raise exceptions here and in the `OSError` below

qmi/core/transport.py
```diff
@@ -686,22 +689,34 @@ def read_until(self, message_terminator: bytes, timeout: Optional[float]) -> byt
...
                 # timeout in socket.recv()
                 nbuf = len(self._read_buffer)
-                raise QMI_TimeoutException("Timeout after {} bytes without message terminator".format(nbuf))
+                raise QMI_TimeoutException(f"Socket timeout after {nbuf} bytes without message terminator")
+            except OSError:
+                raise QMI_RuntimeException(f"UDP packet size was larger than 4096. Data is lost.")
+
+            if self._assert_addr and addr != self._address:
+                _logger.warning(f"Received data from address %s while expected data only from %s!", addr, self._address)
```
rbudhrani, line 713: you can remove the f-string `f` from here too

qmi/core/transport.py
```diff
@@ -723,18 +738,126 @@ def discard_read(self) -> None:
...
+
+    The maximum UDP packet size is 64KB (including headers). For `write` and `read` functions,
+    this means limitations. The write function we can check the data size to send and split into
+    multiple messages. For `read` functions this is not possible and the client must limit the
+    data size as necessary. If for `read` functions the receivable data size is larger than
+    the size defined in `recvfrom(<size>)`, an exception is raised.
+    """
+
+    def __init__(self, host: str, port: int) -> None:
+        """Initialize the TCP transport by connecting to the specified address.
```
rbudhrani, line 770: copy pasta strikes again: TCP -> UDP

qmi/core/transport.py
```diff
@@ -723,18 +738,126 @@ def discard_read(self) -> None:
...
+        # To set the local address to point to our client, we need to bind it. To avoid OSError, we use port number + 1
+        self._socket.bind((self._address[0], self._address[1] + 1))
+
+    def close(self) -> None:
+        _logger.debug("Closing UDP transport %s", self)
+        super().close()
+
+    def write(self, data: bytes) -> None:
+        self._check_is_open()
+        # NOTE: We explicitly adjust the socket timeout before each send/recv call.
```
rbudhrani, line 803: should we mention the reason we do this?

qmi/core/transport.py
```diff
@@ -723,18 +738,126 @@ def discard_read(self) -> None:
...
+        # Create socket and connect.
+        self._socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        self._socket.settimeout(self._connect_timeout)
+        # Set TCP_NODELAY socket option.
+        self._socket.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
+        try:
+            self._socket.connect(self._address)
+        except socket.timeout as e:
+            self._socket.close()
+            raise QMI_TimeoutException("Timeout while connecting to {}".format(self._address)) from e
```
rbudhrani, line 848: can use f-string here

qmi/core/transport.py
```diff
@@ -552,48 +560,38 @@ def discard_read(self) -> None:
...
     @staticmethod
-    def _validate_tcp_port(port):
+    def _validate_port(port):
         if port < 1 or port > 65535:
-            raise QMI_TransportDescriptorException("Invalid TCP port number {}".format(port))
-
-    def __str__(self) -> str:
-        remote_addr = format_address_and_port(self._address)
-        return "QMI_TcpTransport(remote={})".format(remote_addr)
+            raise QMI_TransportDescriptorException(f"Invalid port number {port}")
```
rbudhrani, line 594: this validation is not tested

qmi/core/transport.py
```diff
@@ -613,31 +611,16 @@ def _safe_socket(self) -> socket.socket:
...
-        _logger.debug("Closing TCP transport %s", self)
         super().close()
         self._safe_socket.close()
 
     def write(self, data: bytes) -> None:
-        self._check_is_open()
-        # NOTE: We explicitly adjust the socket timeout before each send/recv call.
-        self._safe_socket.settimeout(None)
-        self._safe_socket.sendall(data)
+        raise NotImplementedError("QMI_UdpTcpTransportBase.write not implemented")
```
rbudhrani, line 623: this is not tested

tests/core/test_transport.py
```diff
@@ -3,16 +3,27 @@
...
+import string
+
+import vxi11
 
 import qmi
 import qmi.core.exceptions
 import qmi.core.transport
-from qmi.core.transport import QMI_TcpTransport, QMI_SerialTransport, QMI_Vxi11Transport, QMI_UsbTmcTransport
+from qmi.core.transport import (
+    QMI_UdpTcpTransportBase,
```
rbudhrani, line 20: can remove this import

qmi/core/transport.py
```diff
@@ -647,18 +630,38 @@ def read(self, nbytes: int, timeout: Optional[float]) -> bytes:
...
+                    # to read. To avoid this happening, read at least 4096 standard that is larger than regular MTU.
+                    udp_packet_size = max(nbytes - nbuf, 4096)
+                    try:
+                        b, addr = self._safe_socket.recvfrom(udp_packet_size)
+                    except (BlockingIOError, socket.timeout) as err:
+                        raise err  # Re-raise this for handling in the upper try-except.
+                    except OSError:
+                        raise QMI_RuntimeException(f"UDP packet size was larger than {udp_packet_size}. Data is lost.")
+
+                    if addr != self._address:
```
rbudhrani, line 644: this branch is not tested

qmi/core/transport.py
```diff
@@ -686,22 +689,34 @@ def read_until(self, message_terminator: bytes, timeout: Optional[float]) -> byt
                 tremain = tstart + timeout - time.monotonic()
                 if tremain < 0:
```
rbudhrani, line 690: this branch is not tested

qmi/core/transport.py
```diff
@@ -686,22 +689,34 @@ def read_until(self, message_terminator: bytes, timeout: Optional[float]) -> byt
...
+                if self._assert_addr:
+                    b, addr = self._safe_socket.recvfrom(4096)
+                else:
+                    b, addr = self._safe_socket.recvfrom(512)
+            except (BlockingIOError, socket.timeout, TimeoutError):
                 # timeout in socket.recv()
                 nbuf = len(self._read_buffer)
-                raise QMI_TimeoutException("Timeout after {} bytes without message terminator".format(nbuf))
+                raise QMI_TimeoutException(f"Socket timeout after {nbuf} bytes without message terminator")
+            except OSError:
```
rbudhrani, line 709: this exception does not seem to have a test

qmi/core/transport.py
```diff
@@ -686,22 +689,34 @@ def read_until(self, message_terminator: bytes, timeout: Optional[float]) -> byt
...
+                raise QMI_RuntimeException(f"UDP packet size was larger than 4096. Data is lost.")
+
+            if self._assert_addr and addr != self._address:
```
rbudhrani, line 712: this branch does not seem to be tested

qmi/core/transport.py
```diff
@@ -723,18 +738,126 @@ def discard_read(self) -> None:
...
             try:
-                b = self._safe_socket.recv(4096)
-            except socket.timeout:
+                b, addr = self._safe_socket.recvfrom(4096)
+            except (BlockingIOError, socket.timeout):
                 # no more bytes available
                 break
-            except BlockingIOError:
-                # no more bytes available
+            except OSError:
```
rbudhrani, line 745: this error and the next if branch need tests

qmi/core/transport.py
```diff
@@ -723,18 +738,126 @@ def discard_read(self) -> None:
...
+        Parameters:
+            host: The server IP address.
+            port: The port for the address.
+        """
+        self._validate_udp_port(port)
+        super().__init__(host, port, True)
+
+    @staticmethod
+    def _validate_udp_port(port):
+        if port == QMI_Context.DEFAULT_UDP_RESPONDER_PORT:
```
rbudhrani, line 781: this condition is not tested

qmi/core/transport.py
```diff
@@ -723,18 +738,126 @@ def discard_read(self) -> None:
...
+            raise QMI_TransportDescriptorException(f"UDP port number {port} not allowed")
+
+    def __str__(self) -> str:
+        remote_addr = format_address_and_port(self._address)
```
rbudhrani, line 785: this magic method doesn't have a test, but I don't know if that's important to test

rbudhrani (changes requested):

qmi/core/transport.py
pieterjbotma, line 644: Do you think this is even possible? In our current implementation. We'll only allow a single transport(socket), to be bound to one UDP device. So wouldn't we expect to only have packets from that address?

pieterjbotma (changes requested):

qmi/core/transport.py
```diff
@@ -613,31 +611,16 @@ def _safe_socket(self) -> socket.socket:
...
-        # Set TCP_NODELAY socket option.
-        self._socket.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
-        try:
-            self._socket.connect(self._address)
-        except socket.timeout as e:
-            self._socket.close()
-            raise QMI_TimeoutException("Timeout while connecting to {}".format(self._address)) from e
-
+        transport_type = "UDP" if self._assert_addr else "TCP"
+        _logger.debug("Connecting %s transport to %s", transport_type, self._address)
```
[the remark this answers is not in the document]
heevasti: Lazy formatting is default for the logger: https://docs.python.org/3/library/logging.html#logging.Formatter. To make sure f-strings would work correctly, we would need to set the format as well.

qmi/core/transport.py
heevasti: Good catch, will be fixed!

qmi/core/transport.py
```diff
@@ -552,48 +560,38 @@ def discard_read(self) -> None:
...
 
+        # As 'localhost' does not necessarily resolve to '127.0.0.1', it is good to resolve it just-in-case.
+        host = socket.gethostbyname(host) if host == "localhost" else host
         self._validate_host(host)
-        self._validate_tcp_port(port)
-
+        self._validate_port(port)
         self._address = (host, port)
-        self._connect_timeout = connect_timeout
+        self._assert_addr = True if assert_address else False  # For UDP, as it accepts data from any client.
```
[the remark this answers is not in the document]
heevasti: Haha, right! Silly me!

qmi/core/transport.py
heevasti: Dayum copy-pasta monster.

qmi/core/transport.py
heevasti: Well this is copied over from the TCP transport and it is there like this. Probably it is so that after a `send` we can call `recv[from]` immediately and it won't give some `OSError: [Errno 98] Address already in use` exception. But I don't know if there is/are some other reason/s.

qmi/core/transport.py
```diff
@@ -552,48 +560,38 @@ def discard_read(self) -> None:
...
 
         Parameters:
-            connect_timeout: Maximum time to connect in seconds.
-
-        Raises:
-            ~qmi.core.exceptions.QMI_TimeoutException: If connecting takes longer than the specified connection
-            timeout.
+            host: The server/client IP address.
+            port: The port for the address.
+            assert_address: Boolean to indicate if the address where data was received from should be asserted.
```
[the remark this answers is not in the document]
heevasti: Going to fix

qmi/core/transport.py
heevasti: Will do

qmi/core/transport.py
heevasti: OK

qmi/core/transport.py
heevasti: OK

qmi/core/transport.py
heevasti: Will remove

qmi/core/transport.py
heevasti: OK

qmi/core/transport.py
heevasti: ADding a test

qmi/core/transport.py
heevasti: ADding a test.

tests/core/test_transport.py
heevasti: I need it now after adding tests for comments above.

qmi/core/transport.py
heevasti: This branch will be removed.

qmi/core/transport.py
heevasti: Added a test for this.

qmi/core/transport.py
heevasti: Added a test

qmi/core/transport.py
heevasti: REmoved this branch

qmi/core/transport.py
heevasti: Added a test.

qmi/core/transport.py
heevasti: Now tested

qmi/core/transport.py
```diff
@@ -647,18 +630,38 @@ def read(self, nbytes: int, timeout: Optional[float]) -> bytes:
         while nbuf < nbytes:
             self._safe_socket.settimeout(tremain)
             try:
-                b = self._safe_socket.recv(nbytes - nbuf)
+                if self._assert_addr:
+                    # For UDP, reading data fails if the arriving packet size is larger than the given nbytes (- nbuf)
+                    # to read. To avoid this happening, read at least 4096 standard that is larger than regular MTU.
+                    udp_packet_size = max(nbytes - nbuf, 4096)
```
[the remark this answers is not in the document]
heevasti: In those cases the client should have some data fragmenting set-up, which divides the large packets into max 4096 bytes. Otherwise, the UDP transport cannot be used for those cases. TCP would work.

qmi/core/transport.py
heevasti: It seemed to give some issues with testing with hardware and the unit-tests after some changes in the transport after hardware tests. So, it is indeed better to remove. In principle it COULD get some packets from elsewhere but I think only if they are sent to this specific port number opened for the UDP. Quite unlikely unless the PC has lots of UDP stuff going on for some reason. 

Removed all these checks.

qmi/core/transport.py
heevasti: For the moment this is necessary to do, as otherwise the Tektronix AWG won't work properly (and possibly same issue could be with other instruments using VXI11 transport, but not seen so far). I would not want to change here the packet size, and changing the method signature is also not possible (Mypy...) as we inherit this from `QMI_Transport` and we should use the same signature.

There is the ticket on our board with name "Rename QMI_Transport read call names and modify their inputs for more logical naming" where we could re-open it for discussion, to see if we could include this. It would help in this case.

qmi/core/transport.py
heevasti: This got removed in the end.

qmi/core/transport.py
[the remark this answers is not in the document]
heevasti: Removed it.

qmi/core/transport.py
```diff
@@ -613,31 +611,16 @@ def _safe_socket(self) -> socket.socket:
...
-        self._socket.settimeout(self._connect_timeout)
-        # Set TCP_NODELAY socket option.
-        self._socket.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
-        try:
-            self._socket.connect(self._address)
-        except socket.timeout as e:
-            self._socket.close()
-            raise QMI_TimeoutException("Timeout while connecting to {}".format(self._address)) from e
-
+        transport_type = "UDP" if self._assert_addr else "TCP"
```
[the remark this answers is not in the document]
heevasti: Removed it.

qmi/core/transport.py
```diff
@@ -647,18 +630,38 @@ def read(self, nbytes: int, timeout: Optional[float]) -> bytes:
         while nbuf < nbytes:
             self._safe_socket.settimeout(tremain)
             try:
-                b = self._safe_socket.recv(nbytes - nbuf)
+                if self._assert_addr:
```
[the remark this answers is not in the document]
heevasti: Removed it.

qmi/core/transport.py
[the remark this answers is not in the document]
heevasti: BTW I noticed different behaviour between unit-tests on my laptop and on the repo runner. Seems like the repo server fragments the packets while locally it does not happen. Could be also a Windows/Linux difference. But the unit-tests should be fine now and work for both.

qmi/core/transport.py
```diff
@@ -612,155 +635,314 @@ def _safe_socket(self) -> socket.socket:
...
-    if hostname[-1] == ".":
-        hostname = hostname[:-1]  # strip exactly one dot from the right, if present
-    parts = hostname.split(".")
-    if re.match(r"^[0-9]+$", parts[-1]):
-        return False  # numeric TLD, not a domain name
-    allowed = re.compile(r"(?!-)[A-Z0-9-]{1,63}(?<!-)$", re.IGNORECASE)
-    return all(allowed.match(x) for x in parts)
+                # Read from socket. NOTE: Reading up to 512 bytes here for TCP,
+                # as larger packet size causes issues with Tektronix 5014 AWG.
+                b, addr = self._read_from_socket(512)
```
pieterjbotma, line 921: major: except for the 512. The rest seems exactly the same. Isn't there a nice way to push this method to the base class. maybe have a MAX_PACKET_SIZE or something constant that changes based on concrete transport.

qmi/core/transport.py
```diff
@@ -612,155 +635,314 @@ def _safe_socket(self) -> socket.socket:
...
+
+        Parameters:
+            host: The server IP address.
+            port: The port for the address.
+        """
+        self._validate_udp_port(port)
+        super().__init__(host, port)
+
+    @staticmethod
+    def _validate_udp_port(port):
```
pieterjbotma, line 741: minor: why not just add this `self._validate_port(...)` method?
```python
    @staticmethod
    def _validate_port(port):
        super()._validate_port(port)
        if port == QMI_Context.DEFAULT_UDP_RESPONDER_PORT:
            raise QMI_TransportDescriptorException(f"UDP port number {port} not allowed")
```

pieterjbotma (changes requested):

qmi/core/transport.py
rbudhrani: sounds good! in that ticket can we link this comment/issue so we don't forget

heevasti: I commented that at the start of the issue and will refine the ticket later on.

qmi/core/transport.py
heevasti: This has been fixed btw.

qmi/core/transport.py
rbudhrani: @pieterjbotma do you know why we set the timeout explicitly?

qmi/core/transport.py
```diff
@@ -612,155 +635,314 @@ def _safe_socket(self) -> socket.socket:
...
 
     def write(self, data: bytes) -> None:
         self._check_is_open()
         # NOTE: We explicitly adjust the socket timeout before each send/recv call.
         self._safe_socket.settimeout(None)
-        self._safe_socket.sendall(data)
+        self._safe_socket.sendto(data, self._address)
 
-    def read(self, nbytes: int, timeout: Optional[float]) -> bytes:
+    def read(self, nbytes: int, timeout: Optional[float] = None) -> bytes:
```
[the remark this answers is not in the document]
heevasti: Almost. There's on L776 (UDP)
`                b, addr = self._read_from_socket(max(nbytes - nbuf, 4096))`
and on L892 (TCP)
`                b, addr = self._read_from_socket(nbytes - nbuf)`. 
The unhandy location within the `while` loop prevented me unifying the methods more. Perhaps it can be done, I just didn't see how to do it without changing the method definition. Adding a new input "packet_size: int = 4096" for example could solve this and it could be then moved into the base class. But I think this is scope creep and should be done in a separate ticket.

qmi/core/transport.py
heevasti: This method could be unified into base class like that. But I would like a solution that works also for the `read` method as well (using it in that one, too). I would push this on another ticket, like the "Rename QMI_Transport read call names and modify their inputs for more logical naming", or make a new ticket altogether.

qmi/core/transport.py
heevasti: Why not? Thanks for the tip, will change this!

qmi/core/transport.py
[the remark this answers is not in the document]
pieterjbotma: If you change the line to something like this, you should be able to push it to the base class.:
```python
    b, addr = self._read_from_socket(max(nbytes - nbuf, MIN_READ_PACKET_LEN))

class UdpTransport(SocketTransport):
    MIN_READ_PACKET_LEN = 4096

class TcpTransport(SocketTransport):
    MIN_READ_PACKET_LEN = 0
```

qmi/core/transport.py
pieterjbotma: Similar to above, making MAX_PACKET_SIZE would work:
```python
class SocketTransport
  def read_untill(...)
    ...
    b, addr = self._read_from_socket(self.MAX_READ_PACKET_LEN )
    ...

class UdpTransport(SocketTransport):
    MAX_READ_PACKET_LEN = 4096

class TcpTransport(SocketTransport):
    MAX_READ_PACKET_LEN = 512
```

pieterjbotma: Do you think it's a good idea to limit our transports based on certain instruments? What if certain instrument can work with larger packet sizes in TCP transport? I feel like it would be better if that packet size could be injected by the instrument itself if it knows it has some limitation.

But like you said, let's make an extra ticket for that case.

qmi/core/transport.py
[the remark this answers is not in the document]
heevasti: OK, so for this `read` we would introduce `MIN_PACKET_SIZE` and for `read_until` (comment below) `MAX_PACKET_SIZE`? Yes, then I guess it could be done. In the base class these would all be set to 0 to make the definition of the class variables so that MyPy won't start heckling.

qmi/core/transport.py
heevasti: Well I think it is necessary for this case if we have an (old) instrument that can handle less than the others. I think 1024 would be better as it is lower than the regular MTU of 1500 and a power of 2 number, but indeed didn't work with the Tektronix. I'll fix this now with the `MAX_PACKET_SIZE = 512` and the suggested (also by @rbudhrani) packet size injection from the instrument will be handled in another ticket. Then for some instruments we could also allow jumbo frames etc.

pieterjbotma (changes requested):

qmi/core/transport.py
```diff
@@ -552,48 +560,70 @@ def discard_read(self) -> None:
...
+    """Base class for bidirectional data streams via socket network connection.
 
-        Raises:
-            ~qmi.core.exceptions.QMI_TimeoutException: If connecting takes longer than the specified connection
-            timeout.
+    Attributes:
+        MIN_PACKET_SIZE: The minimum packet size to read with `read_until` method.
+        MAX_PACKET_SIZE: The maximum packet size to read with `read` method.
+    """
+    MIN_PACKET_SIZE = 0
```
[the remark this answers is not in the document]
heevasti: OK, I'll try.

qmi/core/transport.py
```diff
@@ -612,97 +642,106 @@ def _safe_socket(self) -> socket.socket:
...
-        self._socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-        self._socket.settimeout(self._connect_timeout)
-        # Set TCP_NODELAY socket option.
-        self._socket.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
-        try:
-            self._socket.connect(self._address)
-        except socket.timeout as e:
-            self._socket.close()
-            raise QMI_TimeoutException("Timeout while connecting to {}".format(self._address)) from e
+    def _read_from_socket(self, packet_size: int) -> Tuple[bytes, Any]:
```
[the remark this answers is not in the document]
heevasti: It shortens and cleans up a bit the otherwise messy try-except clauses in the `read` and `read_until`. I like it.

rbudhrani (dismissed):

pieterjbotma (dismissed):

rbudhrani (approved):

PR #64, opened 2024-01-10
Newport single axis motion controller relative move with negative values

rbudhrani:


heevasti (approved):
The change has been covered with a unit-test as well. Ready to merge from my side.

Issue #65, opened 2024-01-11
Adding QMI Drivers for Tenma power supplies


heevasti:
Description 
=========== 
Add QMI drivers for the Tenma benchtop power supply models 72-13360 & 72-2550 in QMI. QMI instrument drivers, already working but rough versions can be found here [13360](https://gitlab.tudelft.nl/team-diamond/diamondos/-/blob/56fc5ca1d0edea3998d995998bd07667df7598e9/instruments/tenma/tenma72_13360.py) and [2550](https://gitlab.tudelft.nl/team-diamond/diamondos/-/blob/56fc5ca1d0edea3998d995998bd07667df7598e9/instruments/tenma/tenma72_2550.py). These are expecting for serial transport strings, but should be edited such that also other transport strings, especially (TBD in [here](https://github.com/orgs/QuTech-Delft/projects/2/views/1?pane=issue&itemId=47294829)) UDP transport can also be used.

There exists already a nice package for controlling Tenma power supplies via serial:
https://github.com/kxtells/tenma-serial/blob/master/tenma/tenmaDcLib.py#L798
From this we could adopt the idea of creating a base class `Tenma_72Base` for our two drivers. In the base class would be defined some class attributes like the max Volt and Amp. We can probably ignore the `NCONFS` as at the moment there is no requirement to save/load configurations. We can leave the default baudrate attribute for serial implementations. Probably (but not checked) some `rpc_method`s could be implemented already in the base class, like `open()`, `close()` and `get_idn()`. For the rest check per method if the implementation can be in the base class. If not, line out the method with inputs, outputs and a docstring, but simply return `NotImplementedError`. The implementations will then be in the "concrete" classes.

Go through the example drivers' code and the pypi code and implement the necessary functions for classes `Tenma_722550` and `Tenma_7213360` in `qmi.instruments.tenma.psu_72` module. Also remember to include description and imports in the `__init__.py` so that the QMI driver show nice description in the Sphinx-generated documentation as well (see other QMI drivers for examples). 

You can also add nice [CLI tools for testing](https://github.com/QuTech-Delft/QMI/tree/main/bin/instruments).

Affected components 
------------------- 
QMI

Modules to be created 
--------------------- 
`qmi/instruments/tenma/psu_72.py`

Modules to be modified
---------------------- 
N/A

Tests to be created/updated 
--------------------------- 
`tests/instruments/tenma/test_psu_72.py`

Documentation to be updated 
--------------------------- 
[...]

PR #66, opened 2024-01-11
Adding qmi drivers for tenma power supplies

heevasti:
Related to #65

Acceptance criteria 
=================== 
- [x] Functionality implemented in agreement with description of the ticket
- [x] Tests are updated
- [x] Documentation updated


qmi/instruments/tenma/psu_72.py
```diff
@@ -0,0 +1,490 @@
...
+            Dictionary of status values.
+        """
+        self._send("STATUS?")
+        statusByte = ord(self._transport.read(1, 1))  # Read response byte
+        ch1mode = (statusByte & 0x01)
+        ch2mode = (statusByte & 0x02)
+        tracking = (statusByte & 0x0C) >> 2
+        output_enabled = (statusByte & 0x40)
+
+        if tracking == 0:
```
rbudhrani, line 256: an enum here would look good, then this whole if block is not needed, instead:
`tracking = TenmaTracking((statusByte & 0x0C) >> 2)` given that `TenmaTracking` is the enum

qmi/instruments/tenma/psu_72.py
```diff
@@ -0,0 +1,490 @@
...
+            tracking = "Independent"
+        elif tracking == 1:
+            tracking = "Tracking Series"
+        elif tracking == 2:
+            tracking = "Tracking Parallel"
+        else:
+            tracking = "Unknown"
+        return {
+            "Ch1Mode": "C.V" if ch1mode else "C.C",
```
rbudhrani, line 265: having an enum for `TenmaChannelMode` with two values `ConstantCurrent` and `ConstantVoltage` would fit well here

qmi/instruments/tenma/psu_72.py
```diff
@@ -0,0 +1,490 @@
...
+    MAX_CURRENT = 30
+    DEFAULT_BAUDRATE = 115200
+
+    def __init__(self, context: QMI_Context, name: str, transport: str) -> None:
+        super().__init__(context, name, transport)
+        self.serial_eol = "\n"
+
+    def _send(self, cmd: str) -> None:
+        cmd = cmd + self.serial_eol
+        return super(Tenma72_13350, self)._send(cmd)
```
rbudhrani, line 337: why do you need arguments for this `super`? Same for other instances

qmi/instruments/tenma/psu_72.py
```diff
@@ -0,0 +1,490 @@
...
+            "OutputEnabled": True | False,
+            "V/C priority ": "Current priority" | "Voltage priority",
+            "Beep ": True | False,
+            "Lock ": True | False,
+
+        Returns: 
+            Dictionary of status values.
+        """
+        self._send("STATUS?")
+        statusBytes = self._transport.read(2, 2)  # Read response byte
```
rbudhrani, line 354: `statusBytes` -> `status_bytes`

qmi/instruments/tenma/psu_72.py
```diff
@@ -0,0 +1,490 @@
...
+        # 72-13350 sends two bytes back, the second being '\n'
+        status = statusBytes[0]
+
+        ch1mode = (status & 0x01)
```
rbudhrani, line 358: `ch1mode` and `current_priority` don't need parenthesis

qmi/instruments/tenma/psu_72.py
```diff
@@ -0,0 +1,490 @@
...
+        output_enabled = bool(status & 0x02)
+        current_priority = (status & 0x04)
+        beep = bool(status & 0x10)
+        lock = bool(status & 0x20)
+
+        return {
+            "ChannelMode": "C.V" if ch1mode else "C.C",
+            "OutputEnabled": output_enabled,
+            "V/C priority": "Current priority" if current_priority else "Voltage priority",
```
rbudhrani, line 367: this can also be an enum. Makes comparisons easier for the layers above QMI

qmi/instruments/tenma/psu_72.py
```diff
@@ -0,0 +1,490 @@
...
+
+    @rpc_method
+    def get_dhcp(self) -> int:
+        """Use the IP LAN command to see if DHCP is enabled.
+
+        Returns:
+            dhcp: The current DHCP enabled state.
+        """
+        cmd = ":SYST:DHCP?" + self.serial_eol
+        super(Tenma72_13350, self)._send(cmd)
```
rbudhrani, line 385: if you do a `self._send(cmd)` then you don't need to make the command as you have in the line before because the `_send` for this class will append the eol. Same for the others

qmi/instruments/tenma/psu_72.py
```diff
@@ -0,0 +1,490 @@
...
+            "OutputEnabled": bool(output_enabled),
+        }
+
+    @rpc_method
+    def enable_output(self, output: bool) -> None:
+        command = f"OUT{int(output)}"
+        self._enable_output(output, command)
+
+
+class Tenma72_2535(Tenma72_2550):
```
rbudhrani, line 277: class dostrings here and for the rest of the inheriting classes would be nice (even if its one line). Maybe you can mention in the docstring what the values of the attributes are, so then our sphinx documentation can also catch it.

bin/instruments/qmi_tenma_72psu.py
```diff
@@ -0,0 +1,150 @@
+#! /usr/bin/env python3
```
rbudhrani, line 1: all our other CLI tools don't have .py appended to their filenames

bin/instruments/qmi_tenma_72psu.py
```diff
@@ -0,0 +1,150 @@
...
+    parser.add_argument("--output", action="store_true", help="Enable output", default=None)
+    parser.add_argument("--no-output", dest="output", action="store_false", help="Disable output")
+    parser.add_argument("--ip", type=str, help="Get/Set IP address", nargs='?', const='get')
+    parser.add_argument("--dhcp", type=str, help="Get/Set IP address", nargs='?', const='get')
+    parser.add_argument("--port", type=str, help="Get/Set IP port number", nargs='?', const='get')
+    parser.add_argument("--mask", type=str, help="Get/Set IP subnet mask", nargs='?', const='get')
+    parser.add_argument("--gateway", type=str, help="Get/Set IP gateway", nargs='?', const='get')
+    args = parser.parse_args()
+
+    instr: globals()[f"Tenma72_{args.model}"]
```
rbudhrani, line 40: i'm not sure how globals works, but I would try with `instr = getattr(qmi.instrument.tenma, f"Tenma_{args.model}")`

qmi/instruments/tenma/psu_72.py
```diff
@@ -0,0 +1,490 @@
...
+    def _enable_output(self, output: bool, command: str) -> None:
+        """Internal function to further communicate the output state to instrument.
+
+        Parameters:
+            output: The expected output state.
+            command: The enable command string.
+        """
+        self._send(command)
+        status = self.get_status()
+        if status["OutputEnabled"] is not output:
```
rbudhrani, line 88: this conditional block does not have a test

qmi/instruments/tenma/psu_72.py
```diff
@@ -0,0 +1,490 @@
...
+        idn = self._read()
+        # Find all relevant info from response. Should be of format TENMA 72-2535 SN:1231345 V2.0
+        pattern = re.compile(r'(\b\w+\b[^ ]*|\b72[^ ]*|V[^ ]*|SN:[^ ]*)')
+
+        # Find all matches in the input string
+        matches = pattern.findall(idn)
+
+        # Extract the relevant parts from the matches
+        vendor = matches[0] if matches else None
+        model = next((match for match in matches[1:] if match.startswith('72')), None)
```
rbudhrani, line 125: would be nice to have a test for the default value of None for these variables

tests/patcher.py
```diff
@@ -0,0 +1,35 @@
+"""Generic patcher implementations."""
+import qmi
```
rbudhrani, line 2: qmi imports should go together

rbudhrani (changes requested):

qmi/instruments/tenma/psu_72.py
```diff
@@ -0,0 +1,490 @@
...
+
+    def _send(self, cmd: str) -> None:
+        """Writes a command to the Tenma, just an ascii string
+
+        Parameters:
+            cmd:    The command to write to the device.
+        """
+        _logger.info("[%s] Sending message %s", self._name, cmd)
+        self._transport.write(bytes(cmd, "ascii"))
+        sleep(.02)  # Sleep some bits to let the message arrive
```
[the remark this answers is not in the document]
heevasti: Made a class attribute

qmi/instruments/tenma/psu_72.py
```diff
@@ -0,0 +1,490 @@
...
+
+    def _read(self, timeout: float = 0.2) -> str:
```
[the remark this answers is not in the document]
heevasti: Changed it.

qmi/instruments/tenma/psu_72.py
```diff
@@ -0,0 +1,490 @@
...
+        """
+        response = self._transport.read_until_timeout(self.BUFFER_SIZE, timeout)
+        _logger.info("[%s] Read message %s", self._name, response)
+        return response.decode('ascii')
+
+    def _enable_output(self, output: bool, command: str) -> None:
+        """Internal function to further communicate the output state to instrument.
+
+        Parameters:
+            output: The expected output state.
```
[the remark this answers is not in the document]
heevasti: Ok

qmi/instruments/tenma/psu_72.py
```diff
@@ -0,0 +1,490 @@
...
+    def read_current(self, output_channel: Optional[int] = None) -> float:
+        """Read the current setting.
+
+        Parameters:
+            output_channel: The channel to send the inquiry to.
+
+        Returns:
+            Current setting of the channel in Amperes.
+        """
+        ch = output_channel or ""
```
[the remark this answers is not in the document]
heevasti: The joke here is that models based on 2550 need a channel number for the command, while based on 13350 not. So when it is `None`, the command is simply "ISET?".  If you make it either mandatory parameter or a check on its value, we would need to make this much more complicated (need to check which class we are in) or make separate implementations for 2550 and 13350 series. 

Now the command will simply not work if sent on 2550 series without a channel number.

qmi/instruments/tenma/psu_72.py
```diff
@@ -0,0 +1,490 @@
...
+        self._send(command)
+        # Response has appended sixth byte from *IDN? to voltage reading due to a firmware bug
+        return float(self._read()[:5])
+
+    @rpc_method
+    def set_current(self, current: float, output_channel: Optional[int] = None) -> None:
+        """Sets the current setting.
+
+        Parameters:
+            current: Current for the channel in Amperes.
```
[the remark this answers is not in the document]
heevasti: I'm afraid I don't see where it is inconsistent.

qmi/instruments/tenma/psu_72.py
```diff
@@ -0,0 +1,490 @@
...
+        command = f"VSET{ch}?"
+        self._send(command)
+        return float(self._read())
+
+    @rpc_method
+    def set_voltage(self, voltage: float, output_channel: Optional[int] = None) -> None:
+        """Sets the voltage setting.
+
+        Parameters:
+            voltage: Voltage setting for the channel in Volts.
```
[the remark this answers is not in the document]
heevasti: Ok

qmi/instruments/tenma/psu_72.py
```diff
@@ -0,0 +1,490 @@
...
+            version=version
+        )
+
+    @rpc_method
+    def get_status(self) -> Dict[str, Any]:
+        """This method needs to be implemented in the inheriting classes."""
+        raise NotImplementedError()
+
+    @rpc_method
+    def read_current(self, output_channel: Optional[int] = None) -> float:
```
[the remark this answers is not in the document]
heevasti: Ok, will change to the usual get-set.

qmi/instruments/tenma/psu_72.py
```diff
@@ -0,0 +1,490 @@
...
+                * 01 = Tracking series
+                * 10 = Tracking parallel
+            OutputEnabled: True | False
+        
+        Returns: 
+            Dictionary of status values.
+        """
+        self._send("STATUS?")
+        statusByte = ord(self._transport.read(1, 1))  # Read response byte
+        ch1mode = (statusByte & 0x01)
```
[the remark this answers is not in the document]
heevasti: removed.

qmi/instruments/tenma/psu_72.py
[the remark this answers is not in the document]
rbudhrani: the tabbing for the docs for `current` and `output_channel` are different

qmi/instruments/tenma/psu_72.py
heevasti: How would that work for the `else: tracking = "Unknown"` part?

rbudhrani: there is a classmethod called `_missing_` that you could override for the enum class.
```python
class TenmaTracking(Enum):
    INDEPENDENT = 0
    TRACKING_SERIES = 1
    TRACKING_PARALLEL = 2
     UNKNOWN = 99999
     
     @classmethod
    def _missing_(cls, value):
        return cls.UNKNOWN
```

qmi/instruments/tenma/psu_72.py
heevasti: Well, my IDE filled it in automatically and I left it. But, in Python 3, the `super()` class knows how to fill these values in automagically, so those arguments are not actually needed. Removing... The result is exactly same.

qmi/instruments/tenma/psu_72.py
heevasti: Whoopsie, correcting.

qmi/instruments/tenma/psu_72.py
heevasti: Removing

qmi/instruments/tenma/psu_72.py
heevasti: Yeah, `super()._send(cmd)` and `self._send(cmd)` are equals in this case as the child class does not implement its own `_send()` method. If it would, then there would be a difference as the first one explicitly calls for the parent class implementation while the second one would call the child class implementation. The order is with `self._send(cmd)`: 1. Is `_send` implemented in this (child) class? Yes -> execute that one or No -> search for it in the parent class and execute there. 

Another difference is that if we decide to implement something different in the sub-sub-class, like in `Tenma72_13360` which is a child of `Tenma72_13350` which is a child of `Tenma72_Base`, using `super()._send()` in the first child would make sure that even if we wrote over the `_send` method there, the base class class method will be called in the second child. And we could overwrite in the second child method the `_send` to use `self._send` with some explicit actions there which would not be affected by the first child `_send`. But, anyhow, we are probably not going to do any of that so I'll go back to the default `self._send` here to avoid any confusion what the idea is. Maybe I had some some initial idea with this to do something else but then it got discarded/forgotten but I left the `super()` stuff here...

heevasti: Oh, sorry, I take this comment back! By replacing it with `self._send(cmd)` results into 
```
  File "C:\Users\heevasti\source\qutech-delft\QMI\qmi\instruments\tenma\psu_72.py", line 338, in _send
    return self._send(cmd)
           ^^^^^^^^^^^^^^^
  [Previous line repeated 976 more times]
RecursionError: maximum recursion depth exceeded
```

so, that was the reason.

qmi/instruments/tenma/psu_72.py
heevasti: Ah, good point! I'll edit those.

qmi/instruments/tenma/psu_72.py
rbudhrani: I think if you change the `_send` method in `Tenma72_13350` to
```python
def _send(self, cmd: str) -> None:
        cmd = cmd + self.serial_eol
        return super()._send(cmd)
```
and then the other calls like `set_dhcp` to
```python
def set_dhcp(self, dhcp: int) -> None:
        """Use the IP LAN command to set a DHCP enabled state.

        Parameters:
            dhcp: New DHCP state. 0 is disabled, 1 is enabled.
        """
        cmd = f":SYST:DHCP {dhcp}" + self.serial_eol
        self._send(cmd)
```
it will fix that issue

qmi/instruments/tenma/psu_72.py
rbudhrani: I realised with the other comment that I skimmed over the classname. This is what causes the recursions, because you are calling the `_send` method of `Tenma72_13350` inside its own `_send` method. I'm surprised `get_status` did not complain with the recursion error

qmi/instruments/tenma/psu_72.py
heevasti: OK, my idea was to use the specific implementations of the `_send(self, cmd)` method in this (TEnma72_13350) class to not to have to write `cmd = "come_command" + self.serial_eol` for each command. But as I DID write it that way, calling `self._send(cmd)` made it to add it twice (`self.serial_eol`) and of course it failed. The correct fix is to keep the `_send` implementation as-is in this class, but replacing the `super().send` with `self.send`, and removing the ` + self.serial_eol` in all the RPC methods. Good that you dug into this so that I realized what the "issue" (well, sloppy implementation that happened to also work) was.

rbudhrani: did the code snippet I commented work? I'm surprised that writing it that way still gives recursion issues 🤔

bin/instruments/qmi_tenma_72psu.py
heevasti: Good catch, I'll remove the extension!

bin/instruments/qmi_tenma_72psu.py
heevasti: Both work, and give the correct instance back. But I suppose the `getattr` version is safer as it limits the search in `qmi.instruments.tenma` package, and not ALL globals. Even though I really doubt there would be some other global variable called `Tenma_xxx`...

qmi/instruments/tenma/psu_72.py
heevasti: Hmm... For `vendor`, `model`, `version` nothing (exciting) happens, but indeed `serial` would cause an `AttributeError` in the `serial=serial.strip()[3:],`. I think a check on this is needed to return `None` instead of excepting on it. Then just returning `None` for all `QMI_InstrumentIdentification` values is a reasonable test.

heevasti: ... unless it would be preferable to raise an exception in that case?

tests/patcher.py
heevasti: Fixed.

qmi/instruments/tenma/psu_72.py
heevasti: The client seems to prefer values "C.V" and "C.C". Those cannot be made with `Enum` as it complains about the `.` in the name.

qmi/instruments/tenma/psu_72.py
```diff
@@ -0,0 +1,490 @@
...
+
+        if tracking == 0:
+            tracking = "Independent"
+        elif tracking == 1:
+            tracking = "Tracking Series"
+        elif tracking == 2:
+            tracking = "Tracking Parallel"
+        else:
+            tracking = "Unknown"
+        return {
```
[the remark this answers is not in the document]
heevasti: Yes, but I don't like to do it because the other `get_status` of `Tenma72_13350` does have different outputs. So, it would get complicated as I would need to make two `NamedTuple`s (or `dataclass`es) as make the base class to return an `Union` of those. The dictionary as `Dict[str, Any]` works for both so makes it simpler.

qmi/instruments/tenma/psu_72.py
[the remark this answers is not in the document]
heevasti: Oh right, I looked in the `get` method. Sorry.

qmi/instruments/tenma/psu_72.py
heevasti: OK, I will make this as an Enum. With more than 2 possible values it makes sense.

qmi/instruments/tenma/psu_72.py
heevasti: Enum just for two values does not make much of a difference. And I think the client wants to use the string as output so I would prefer to leave it.

qmi/instruments/tenma/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+  - 72-2535, 72-2540, 72-2545, 72-2550, 72-2925, 72-2930, 72-2935, 72-2940 & 72-10480
+  - 72-13350 & 72-13360
+"""
+import importlib
+
+tenma_module = importlib.import_module("qmi.instruments.tenma.psu_72")
+psu_classes = {name: cls for name, cls in dict(vars(tenma_module)).items() if name.startswith("Tenma")}
+for k, v in psu_classes.items():
+    if "Base" not in k:
+        globals()[k] = v
```
rbudhrani, line 15: do you need to add to globals now that you use `getattr`

qmi/instruments/tenma/psu_72.py
rbudhrani: I'm happy with returning None, but this scenario is not tested yet

qmi/instruments/tenma/psu_72.py
rbudhrani:
```python
class TenmaChannelMode(StrEnum):
    CONSTANT_CURRENT = "C.C"
    CONSTANT_POWER = "C.V"
```
this should work

qmi/instruments/tenma/__init__.py
heevasti: Looks like it. If I remove this, the script CLI raises an error in the `getattr` line.

qmi/instruments/tenma/psu_72.py
heevasti: Well, the snippet you wrote had still that ` + self.serial_eol` in the RPC method, which made it add the EOL char twice. So removing that from the RPC calls and having it only in the `_send` works.

qmi/instruments/tenma/psu_72.py
heevasti: Well I made it to raise an exception, if that's OK.

qmi/instruments/tenma/psu_72.py
heevasti: But then I would have to implement that line as:
```python
    "Ch1Mode": TenmaChannelMode.CONSTANT_CURRENT if ch1mode else TenmaChannelMode.CONSTANT_POWER [sic],
``` 
It doesn't really make it nicer.

rbudhrani: from enum import Enum
class TenmaChannelMode(Enum):
    CONSTANT_CURRENT = 0
    CONSTANT_POWER = 1

qmi/instruments/tenma/psu_72.py
rbudhrani: addressed in other comment

qmi/instruments/tenma/__init__.py
rbudhrani: fixed with the new import statement

qmi/instruments/tenma/psu_72.py
heevasti: OK, I made a dataclass for it. See latest commit.

qmi/instruments/tenma/psu_72.py
```diff
@@ -0,0 +1,562 @@
...
+    value: bool
+    name: str = "C.C"
+
+    def __post_init__(self):
+        if self.value:
+            self.name = "C.V"
+
+
+class TrackingState(Enum):
+    Independent = 0
```
rbudhrani, line 43: Enum values should be in all caps

qmi/instruments/tenma/psu_72.py
rbudhrani: if its a dataclass that returns string, then the ternary statement you had earlier also works

rbudhrani (approved):

PR #73, opened 2024-03-15
62 new qmi driver for thorlabs mpc320

rbudhrani:


qmi/instruments/thorlabs/mpc320.py
```diff
@@ -0,0 +1,507 @@
...
+    AptMessageId,
+    AptProtocol,
+)
+
+# Global variable holding the logger for this module.
+_logger = logging.getLogger(__name__)
+
+
+@dataclass
+class Thorlabs_MPC320_Status:
```
heevasti, line 33: Minor: For these (data)classes we do not apply the QMI-driver naming convention, so you could use the regular 'CamelCase' naming, without underscores.

qmi/instruments/thorlabs/mpc320.py
```diff
@@ -0,0 +1,507 @@
...
+
+    # the maximum range for a paddle is 170 degrees
+    # the value returned by the encoder is 1370 for 170 degrees
+    ENCODER_CONVERSION_UNIT = 170 / 1370
+
+    MIN_POSITION_DEGREES = 0
+    MAX_POSITION_DEGREES = 170
+
+    MIN_VELOCITY_PERC = 10
+    MAX_VELOCITY_PERC = 100
```
heevasti, line 88: You are using `int`s here, could they also be floats?

qmi/instruments/thorlabs/apt_protocol.py
```diff
@@ -0,0 +1,249 @@
...
+        self._transport = transport
+        self._timeout = default_timeout
+        self._apt_device_address = apt_device_address
+        self._host_address = host_address
+
+    def write_param_command(
+        self,
+        message_id: int,
+        param1: Optional[int] = None,
+        param2: Optional[int] = None,
```
heevasti, line 184: Couldn't you just do `int = 0`? Then in the message you just use `param1`, `param2` as 0x00 == 0.

qmi/instruments/thorlabs/apt_protocol.py
```diff
@@ -0,0 +1,249 @@
...
+
+class AptMessageHeaderForData(LittleEndianStructure):
+    """
+    This is the version of the APT message header when a data packet follows a header.
+    """
+
+    _pack_ = True
+    _fields_: List[Tuple[str, type]] = [
+        ("message_id", c_uint16),
+        ("date_length", c_uint16),
```
heevasti, line 131: I suppose you should refer here to message data length, not message date as in day,month,year...?

qmi/instruments/thorlabs/mpc320.py
```diff
@@ -0,0 +1,507 @@
...
+    def _validate_position(self, pos: float) -> None:
+        """
+        Validate the position. Any position for the MPC320 needs to be in the range 0 to 170 degrees,
+        or 0 to 1370 in encoder counts.
+
+        Parameters:
+            pos:    Position to validate in degrees.
+
+        Raises:
+            an instance of QMI_InstrumentException if the position is invalid.
```
heevasti, line 113: Minor: In the readthedocs this looks nicer if you do `QMI_InstrumentException: If the position is invalid`. The `:` makes sure the exception name is in bold and the description with regular lettering. Valid to all `Raises:` sections in these docstrings.

qmi/instruments/thorlabs/mpc320.py
```diff
@@ -0,0 +1,507 @@
...
+        Validate the channel number. The MPC320 has 3 channels.
+
+        Parameters:
+            channel_number: Channel number to validate.
+
+        Raises:
+            an instance of QMI_InstrumentException if the channel is not 1, 2 or 3
+        """
+
+        if channel_number not in [1, 2, 3]:
```
heevasti, line 149: Minor: If you want this to be more flexible, you could do `range(self.MIN_CHANNEL_NUMBER, self.MAX_CHANNEL_NUMBER + 1)`. So it can be directly applied also for 2 channel cases, if that gets implemented.

qmi/instruments/thorlabs/mpc320.py
```diff
@@ -0,0 +1,507 @@
...
+        and fail to include it/them again in this call, that channel(s) will be disabled. For example, if you run the
+        following:
+        self.enable_channel([1])
+        self.enable_channel([2])
+        only channel 2 will be enabled and 1 and 3 will be disabled.
+        The correct way to call this method in this case is
+        self.enable_channel([1,2])
+
+        Parameters:
+            channel_number: The channnels(s) to enable.
```
heevasti, line 210: triple-n

qmi/instruments/thorlabs/mpc320.py
```diff
@@ -0,0 +1,507 @@
...
+        # For the MPC320 the channel number does not matter here. The device has one LED that flashes irrespective
+        # of the provided channel number.
+        self._apt_protocol.write_param_command(AptMessageId.MOD_IDENTIFY.value, 0x01)
+
+    @rpc_method
+    def enable_channels(self, channel_numbers: List[int]) -> None:
+        """
+        Enable the channel(s). Note that this method will disable any channel that is not provided as an argument. For
+        example, if you enable channel 1, then 2 and 3 will be disabled. If you have previously enabled a channel(s)
+        and fail to include it/them again in this call, that channel(s) will be disabled. For example, if you run the
```
heevasti, line 201: This must be the Thorlabs logic for this function? As I don't see anywhere a 'disable' list.

qmi/instruments/thorlabs/mpc320.py
```diff
@@ -0,0 +1,507 @@
...
+    @rpc_method
+    def is_channel_homed(self, channel_number: int, timeout: float = DEFAULT_RESPONSE_TIMEOUT) -> bool:
+        """
+        Check if a given channel is homed. This command should only be run after the method `home_channel`.
+        Otherwise you will read bytes from other commands using this method.
+
+        Paramters:
+            channel_number: The channel to check.
+            timeout:        The time to wait for a response to the homing command. This is optional
+                            and is set to a default value of DEFAULT_RESPONSE_TIMEOUT.
```
heevasti, line 311: Reformat to 'with default value DEFAULT_RESPONSE_TIMEOUT'.

qmi/instruments/thorlabs/mpc320.py
```diff
@@ -0,0 +1,507 @@
...
+        # then read the actual response we need
+        resp = self._apt_protocol.ask(MOT_MOVE_HOMED, timeout)
+
+        # Check if the channel number in the response is equal to the one that was asked for.
+        return resp.chan_ident == Thorlabs_MPC320_ChannelMap[channel_number]
+
+    @rpc_method
+    def move_absolute(self, channel_number: int, position: float) -> None:
+        """
+        Move a channel to the specified position. The specified position is in degeres. A conversion is done to convert
```
heevasti, line 331: degeres -> degrees

qmi/instruments/thorlabs/mpc320.py
```diff
@@ -0,0 +1,507 @@
...
+        self._validate_channel(channel_number)
+        self._check_is_open()
+        # This command needs a workaround. Instead of returning the MOT_MOVE_HOMED message, the instrument returns
+        # its current state first, so read that to discard the buffer
+        _ = self._apt_protocol.ask(MOT_GET_USTATUSUPDATE, timeout)
+        # then read the actual response we need
+        resp = self._apt_protocol.ask(MOT_MOVE_HOMED, timeout)
+
+        # Check if the channel number in the response is equal to the one that was asked for.
+        return resp.chan_ident == Thorlabs_MPC320_ChannelMap[channel_number]
```
heevasti, line 326: Question: I find this a little baffling (probably a Thorlabs bug): So the response is sent only when the move is completed and the returned "chan_ident" just says the channel number? This could receive an answer of another channel and fail, right?

qmi/instruments/thorlabs/mpc320.py
```diff
@@ -0,0 +1,507 @@
...
+    def jog(
+        self,
+        channel_number: int,
+        direction: AptChannelJogDirection = AptChannelJogDirection.FORWARD,
+    ) -> None:
+        """
+        Move a channel specified by its jog step.
+
+        Parameters:
+            channel_number: The channel to job.
```
heevasti, line 439: job -> jog. Also next line

qmi/instruments/thorlabs/mpc320.py
```diff
@@ -0,0 +1,507 @@
...
+            jog_step3=round(jog_step3 / self.ENCODER_CONVERSION_UNIT),
+        )
+        # Send message.
+        self._apt_protocol.write_data_command(AptMessageId.POL_SET_PARAMS.value, data_packet)
+
+    @rpc_method
+    def get_polarisation_parameters(self) -> Thorlabs_MPC320_PolarisationParameters:
+        """
+        Get the polarisation parameters.
+        """
```
heevasti, line 494: Returns: ...

heevasti (commented):
Great work here! Some comments and questions. Don't forget to add the shortcut to the class in `__init__.py` and edit the `CHANGELOG.md`. Awaiting for the unit-tests.

The protocol seems fine to me in general.

tests/instruments/thorlabs/test_mpc320.py
```diff
@@ -0,0 +1,537 @@
...
+    AptChannelState,
+)
+
+
+class TestThorlabsMPC320(unittest.TestCase):
+    def setUp(self):
+        # Patch QMI context and make instrument
+        self._ctx_qmi_id = f"test-tasks-{random.randint(0, 100)}"
+        qmi.start(self._ctx_qmi_id)
+        self._transport_mock = unittest.mock.MagicMock(spec=QMI_SerialTransport)
```
heevasti, line 19: So, there was this discussion of not using the `qmi.start` in unit-tests anymore. It is indeed unnecessary to open localhost contexts for unit-tests as we actually just need to test the driver. To replace this you can use the `/tests/patcher.py`.

You can either patch the `qmi.start` with
```python
from tests.patcher import PatcherQmiContext
...
        self._ctx_qmi_id = f"test-tasks-{random.randint(0, 100)}"
        self.qmi_patcher = PatcherQmiContext()
        self.qmi_patcher.start(self._ctx_qmi_id)
```
which is patching for this type of tests. For testing the driver class directly you can do
```python
from tests.patcher import PatcherQmiContext as QMI_Context
...
        patcher = patch('qmi.instruments.thorlabs.mpc320.create_transport', spec=QMI_SerialTransport)
        self._transport_mock = patcher.start().return_value
        self.addCleanup(patcher.stop)
        # Make DUT
        self._instr: Thorlabs_Mpc320 = Thorlabs_Mpc320(
            QMI_Context("test_mpc320"), Thorlabs_Mpc320, "serial:transport_str"
        )
        self._instr.open()
```

qmi/instruments/thorlabs/__init__.py
```diff
@@ -7,8 +7,11 @@
 - PM100D power meter
 - TSP01, TSP01B environmental sensors.
 """
```
heevasti, line 9: Add MPC320 into this list

qmi/instruments/thorlabs/apt_protocol.py
rbudhrani: indeed!

qmi/instruments/thorlabs/apt_protocol.py
rbudhrani: yes! good catch

qmi/instruments/thorlabs/apt_packets.py
```diff
@@ -0,0 +1,186 @@
...
+        msg_id:     ID of message whose settings should be save.
+    """
+
+    MESSAGE_ID = AptMessageId.MOT_SET_EEPROMPARAMS.value
+    _fields_: List[Tuple[str, type]] = [("chan_ident", apt_word), ("msg_id", apt_word)]
+
+
+class POL_GET_SET_PARAMS(AptMessage):
+    """ "
+    Data packet structure for POL_SET_PARAMS command. It is also the data packet structure for the POL_SET_PARAMS.
```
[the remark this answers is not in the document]
rbudhrani: good catch! the second one is now GET

qmi/instruments/thorlabs/mpc320.py
rbudhrani: for this I want to start using the instrument as a prefix because then the client will know what its for. In general I want to start this in other QMI drivers as well. let me know what you think?

qmi/instruments/thorlabs/mpc320.py
rbudhrani: also the double s 😄

qmi/instruments/thorlabs/mpc320.py
rbudhrani: yes, for this driver, you automatically disable channels that are not in the enable list. That's why for the `disable_all_channels()` command you will notice that I use the same message ID `MOD_SET_CHANENABLESTATE` but with an empty list. So I am enabling no channels, which automatically disables all of them

qmi/instruments/thorlabs/mpc320.py
rbudhrani: they can only be ints

tests/instruments/thorlabs/test_mpc320.py
rbudhrani: was meaning to change that, thanks for pointing it out. I have still kept the patcher for the transport with the `with` keyword

qmi/instruments/thorlabs/mpc320.py
rbudhrani: nice! did it for all

qmi/instruments/thorlabs/mpc320.py
heevasti: I have no issue with that. The linter probably does complain about it though?

rbudhrani: linter seems happy with it

qmi/instruments/thorlabs/mpc320.py
rbudhrani: This was the case for `MOD_GET_CHANENABLESTATE` so I have do it for the others as well.
The response is only sent when the move is completed, which is why I now have a try except block. If it receives the info for another channel I would like it to fail, since I don't know whether the channel that I wanted info for is homed or not

heevasti (approved):

PR #75, opened 2024-03-18
Fixing pipelines

rbudhrani:


qmi/instruments/teraxion/tfn.py
```diff
@@ -399,8 +414,8 @@ def get_manufacturer_name(self) -> str:
         resp = self._read(Teraxion_TFNCommand_GetManufacturerName)
         time.sleep(self.GET_LONG_PROCESS_TIME)
         # get the data after the status bytes and ignore the null bytes
-        manufacturer_name = resp[self.LEN_STATUS_BYTES:]
-        return manufacturer_name.decode("ascii")[:manufacturer_name.find(b"\x00")]
+        manufacturer_name = resp[self.LEN_STATUS_BYTES :]
+        return manufacturer_name.decode("ascii")[: manufacturer_name.find(b"\x00")]
```
heevasti, line 418: Minor: These spaces around the `:` look a bit ugly. Is it Black that's doing that? Anyhow, a preference thing I suppose.

setup.py
```diff
@@ -9,114 +9,109 @@
...
             "astroid==2.12.2",
             "coverage",
-            'pylint',
-            'mypy',
-            "radon",
+            "pylint",
+            "mypy",
             # version bump
-            'bump2version'
+            "bump2version",
```
heevasti, line 86: We could remove `bump2version` in the OSS QMI, as we don't use it here. The releases and versioning is done manually. We can put it back if we get a procedure for releasing using it.

rbudhrani: works for me!

heevasti (dismissed):

heevasti (approved):

PR #78, opened 2024-05-27
77 update zhinstzipython package to zhinstcore

rbudhrani:


rbudhrani:
A new version of the driver using `zhinst-toolkit` instead of `zhinst-core` is the new scope of this ticket.

pieterjbotma (approved):

Issue #83, opened 2024-07-10
Limit NumPy version to <2.0 until support for p38 is dropped


heevasti:
Description
-------------
The new NumPy major release (2.0) has many breaking changes. And it also does not support officially Python 3.8 anymore. Therefore, until [ticket](https://github.com/orgs/QuTech-Delft/projects/2/views/1?pane=issue&itemId=43902246) is to be executed, we should limit the NumPy dependency to versions under 2.0 in `setup.py`. Probably we could do the same for SciPy just in case, if they are planning to make a major upgrade as well.

Affected components
------------------------
QMI

Modules to be modified
---------------------------
`setup.py`

Documentation to be updated
----------------------------------
`CHANGELOG.md`

PR #84, opened 2024-07-10
[QMI-083] Limit NumPy and SciPy versions <2.

heevasti:
To solve ticket #83

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,8 +56,8 @@ setup(
         "wheel",
         "twine",
         # For scientific data processing and visualisation
-        "numpy",
-        "scipy",
+        "numpy<2",
+        "scipy<2",
         "h5py>=3.7.0",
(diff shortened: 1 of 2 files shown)


pieterbuzing (approved):

setup.py
```diff
@@ -56,8 +56,8 @@ def read(*names, **kwargs):
         "wheel",
         "twine",
         # For scientific data processing and visualisation
-        "numpy",
-        "scipy",
+        "numpy<2",
+        "scipy<2",
```
rbudhrani, line 60: the latest version of scipy is still 1.14

heevasti: I suppose SciPy will come with an `2.0` soonish as well, following `NumPy`. Just making it future proof.

rbudhrani: to be fixed later

PR #85, opened 2024-07-12
[QMI-079] Created a context manager...

heevasti:
…  for entering and exiting configuration mode and refactored the methods using the configuration mode.

### DIFF
--- a/qmi/instruments/newport/single_axis_motion_controller.py
+++ b/qmi/instruments/newport/single_axis_motion_controller.py
@@ -2,6 +2,7 @@
 Instrument driver for a Newport single axis motion controller. This is a base class for other controllers,
 but can be used without extending.
 """
+from contextlib import contextmanager
 import enum
 import logging
 from time import sleep
@@ -138,6 +139,13 @@ class Newport_SingleAxisMotionController(QMI_Instrument):
             raise QMI_InstrumentException(
                 f"Invalid controller address {self._controller_address}")
 
+    @contextmanager
+    def configuration_state(self):
+        try:
+            yield self._enter_configuration_state()
+        finally:
+            self._exit_configuration_state()
+
     def _build_command(
             self, cmd: str, value: Union[float, int, None] = None
     ) -> str:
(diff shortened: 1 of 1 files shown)


qmi/instruments/newport/single_axis_motion_controller.py
```diff
@@ -138,6 +139,13 @@ def controller_address(self, controller_address: Optional[int]) -> None:
             raise QMI_InstrumentException(
                 f"Invalid controller address {self._controller_address}")
 
+    @contextmanager
```
rbudhrani, line 142: Can we unittest the context manager separately?

qmi/instruments/newport/single_axis_motion_controller.py
```diff
@@ -622,22 +641,23 @@ def get_acceleration(self, controller_address: Optional[int] = None) -> float:
...
+            acceleration = self._scpi_protocol.ask(
+                self._build_command("AC?"))
+            sleep(self.COMMAND_EXEC_TIME)
+            self._check_error()
+            return acceleration
+
         _logger.info("Getting acceleration of instrument [%s]", self._name)
         self.controller_address = controller_address
         # instrument must be in CONFIGURATION, DISABLE or READY state to get the acceleration.
         try:
```
rbudhrani, line 654: I see that in a lot of commands we do this `try` `catch` to check if the instrument is already in the ready state. Perhaps we can do this in the context manager. Then we can validate the ready state in the context manager instead of repeating the check for each method that needs it. The commands would like something like:
```python
with self.configuration_state("acceleration"):
    acceleration = _get_acceleration()
```

rbudhrani: also then you don't need a private method like `_get_acceleration()`, because I think you made it so you could use the same in the `try` and `catch` block.

heevasti: I think this would not work. This is because we use the `try`-`except` block to decide if we need to use the context manager. Moving the check inside the context manager defeats its purpose as we will then always be using the context manager even when it is not needed.

qmi/instruments/newport/single_axis_motion_controller.py
heevasti: Coverage reports 100% so I don't think there's a need for it. Perhaps if it would be a more complex one with inputs and/or logic, but now it is a very simple one so no.

rbudhrani (approved):

PR #95, opened 2024-08-09
[QMI-094]

heevasti:
Correcting and improving documentation across QMI transport related files. Also made some reformatting and typing changes, namely replacing `Mapping` with `Dict` when generic `Mapping` type is unnecessarily vague. And `Mapping` is imported now from `abc.collections` due to obsoletion in `typing`.


qmi/core/transport.py
```diff
@@ -179,16 +182,30 @@ def discard_read(self) -> None:
...
+                 keywords: Mapping[str, Tuple[Type, bool]]
+    ) -> None:
         self.interface = interface
         self._positionals = positionals
         self._keywords = keywords
 
-    def parse_parameter_strings(self, transport_descriptor: str, default_parameters=None) -> Mapping[str, Any]:
+    def parse_parameter_strings(
+            self, transport_descriptor: str, default_parameters: Optional[Dict[str, Any]] = None
+    ) -> Mapping[str, Any]:
```
rbudhrani, line 199: the return type here would also be a `Dict`

qmi/core/transport.py
```diff
@@ -262,7 +280,7 @@ def _parse_positional_parameters(self, params: List[str]) -> Mapping[str, Any]:
                                                        ty, param)
         return d
 
-    def _parse_keyword_parameters(self, strings: List[str]) -> Mapping[str, Any]:
+    def _parse_keyword_parameters(self, strings: List[str]) -> Dict[str, Any]:
```
rbudhrani, line 283: in this method there's a type check that I cannot comment on. Its line 294:
```python
if ty == int ...
```
this can be changed to `if ty is int`
Only if you want to boyscout a bit more

qmi/core/transport.py
heevasti: If Mypy does not complain, it should be fine like this. I'll leave it for now as 'original' as I get no trouble with it.

qmi/core/transport.py
heevasti: OK, changing it, why not :)

qmi/core/transport.py
rbudhrani: hmmm, with this typing we are implying 2 things:
1. somewhere in this function we have changed the `parameters` from `Dict` to `Mapping` (as we have changed the typing for the other private methods from `Mapping` to `Dict`)
2. `Mapping` is immutable and `Dict` is not, so the parameters we return are immutable according to our function signature

heevasti: I see, that is misleading indeed. It can be changed to `MutableMapping` to solve this issue.

rbudhrani: I think that will work. `Dict` inherits from `MutableMapping` (which is an ABC), but for typing I don't think that should make a difference. But to be safe and correct I still prefer `Dict`, because inside this method `parameters` is always of type `Dict`

heevasti: OK, I'll change it.

qmi/core/transport.py
```diff
@@ -179,16 +182,30 @@ def discard_read(self) -> None:
...
 class TransportDescriptorParser:
-
+    """This class is for creating a transport-specific parser classes and has (static) methods that are
+    used for parsing transport strings.
+    """
     def __init__(self,
                  interface: str,
                  positionals: List[Tuple[str, Tuple[Type, bool]]],
-                 keywords: Mapping[str, Tuple[Type, bool]]):
+                 keywords: MutableMapping[str, Tuple[Type, bool]]
```
rbudhrani, line 191: i missed this one, but if we want to get rid of `MutableMapping` this is the last place to change it. Then we don't need to import it either

heevasti: From https://docs.python.org/3/library/typing.html#typing.Dict, (`Mutable`)`Mapping` is preferred for input arguments, Thus I would like change this back to the original `Mapping`, as we do not need to mutate this input. The same goes for the `default_parameters` input in `parse_parameter_strings`.

heevasti: The changes for output types to `Dict` we can thus keep.

rbudhrani (approved):

PR #98, opened 2024-08-22
90 add   enter   and   exit   methods to qmi instrument

heevasti:
- The `QMI_Instrument` and `QMI_TaskRunner` (which inherit from `QMI_RpcObject`) are now equipped with specific `__enter__` and `__exit__` methods, which in the case of `QMI_Instrument`
  also open and close the instrument when run with a `with` context manager protocol. Meanwhile `QMI_TaskRunner` starts and stops then joins a QMI task thread. In practise, these context managers
  can be used instead of the to-be-obsoleted `open_close` and `start_stop_join` context managers. The context manager protocol cannot be used for `QMI_RpcObject` directly.

- Relevant unit-tests added
- Docstrings added and improved.

- Fix on scheduled full CI unit-test pipeline which was missing installing of dependencies.

### DIFF
--- a/.github/workflows/scheduled-full-ci.yml
+++ b/.github/workflows/scheduled-full-ci.yml
@@ -29,6 +29,7 @@ jobs:
       - name: Run unit tests and generate report
         if: always()
         run: |
+          pip install -e .
           pip install unittest-xml-reporting
           python -m xmlrunner --output-file testresults.xml discover --start-directory=tests --pattern="test_*.py"
 
--- a/.github/workflows/scheduled-full-ci.yml
+++ b/.github/workflows/scheduled-full-ci.yml
@@ -29,7 +29,7 @@ jobs:
       - name: Run unit tests and generate report
         if: always()
         run: |
-          pip install -e .
+          pip install .
           pip install unittest-xml-reporting
           python -m xmlrunner --output-file testresults.xml discover --start-directory=tests --pattern="test_*.py"
 
--- a/qmi/core/rpc.py
+++ b/qmi/core/rpc.py
@@ -83,7 +83,7 @@ The normal proxy locking is available only within the same context for the same 
 Since QMI V0.29.1 it is also possible to lock and unlock with a custom token from
 other proxies as well, as long as the contexts for the other proxies have the same name.
 
-Example 1:
+Example 1::
         # Two proxies in same context.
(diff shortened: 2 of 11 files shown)


.github/workflows/scheduled-full-ci.yml
```diff
@@ -29,6 +29,7 @@ jobs:
       - name: Run unit tests and generate report
         if: always()
         run: |
+          pip install -e .
```
pieterjbotma, line 32: minor, but is `-e` really necessary? I think that's mainly when you want keep developing code in that package. the runner just needs to install and use.

heevasti: Didn't really think this, I just copied it over. I suppose removing it is better, although I suspect minor difference in running the pipelines.

qmi/core/rpc.py
```diff
@@ -659,6 +676,26 @@ def make_rpc_forward_function(method_name: str):
         # Add non-blocking proxy.
         self.rpc_nonblocking = QMI_RpcNonBlockingProxy(context, descriptor)
 
+    def __enter__(self) -> "QMI_RpcProxy":
+        """The context manager definition is needed for the proxy as it will always be returned from QMI contexts,
```
pieterjbotma, line 680: I think it's probably a good idea to put [this](https://docs.python.org/3/reference/datamodel.html#special-method-lookup) link in here for people who'd like to deep-dive.

heevasti: I still don't really get how that link relates with the Recursion error?

rbudhrani: It explains why we need `__enter__` and `__exit__` in the `QMI_RpcProxy` class. Normally, you expect that having it in `QMI_Instrument` would suffice, but because we return a `QMI_RpcProxy` we have to implement the context management protocol in `QMI_RpcProxy` (just as a stub) to satisfy the `with` keyword. The `with` keyword looks in the class to see if the context management protocol is implemented and not in the instance.

pieterjbotma: with _actually executes_ the type(obj).__enter__ and type(obj).__exit__ methods.

pieterjbotma: those class methods in turn call the instance methods, which are rpc methods

pieterjbotma (approved):

PR #103, opened 2024-09-20
102 import new bristol fos driver from diamondos

heevasti:
Added the Windows version of the Bristol FOS QMI driver and CLI client for it. Unit-tests also updated.

### DIFF
--- a/bin/instruments/qmi_bristol_fos
+++ b/bin/instruments/qmi_bristol_fos
@@ -1,14 +1,13 @@
 #! /usr/bin/env python3
 
 """Command line client for the Bristol FOS Fiber-optic switch."""
-
-import argparse
-from contextlib import nullcontext, AbstractContextManager
 import sys
+import argparse
+from contextlib import nullcontext, AbstractContextManager, ExitStack
 
 import qmi
 from qmi.instruments.bristol import Bristol_Fos
-from qmi.utils.context_managers import start_stop, open_close
+from qmi.utils.context_managers import start_stop
 
 
 def main() -> None:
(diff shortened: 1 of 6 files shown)


bin/instruments/qmi_bristol_fos
```diff
@@ -0,0 +1,64 @@
...
+    source.add_argument("--rpc", help="QMI RPC address of device.")
+    source.add_argument("--unique_id", help="Unique ID of the device")
+
+    parser.add_argument("--board_id", type=int, help="Board number for the device. Windows only!")
+    parser.add_argument("--channel", type=int, help="Channel number to set")
+
+    args = parser.parse_args()
+
+    instr: Bristol_Fos
+    with start_stop(qmi, "bristol_fos_client", console_loglevel="WARNING"), parse_source(args) as instr:
```
rbudhrani, line 29: if we don't chain the `with`s, then we can still use the built-in context management protocol here:
```python
with <qmi_context_stuff>:
   if args.unique_id is not None and args.board_id is not None:
       with qmi.make_instrument(
                instrument_name="FOS",
                instrument_class=Bristol_Fos,
                unique_id=args.unique_id,
                board_id=int(args.board_id) as instr:
            instr.select_channel(args.channel)
    elif args.unique_id is not None:
        qmi.make_instrument(
                instrument_name="FOS",
                instrument_class=Bristol_Fos,
                unique_id=args.unique_id
        ) as instr:
            instr.select_channel(args.channel)
... # any other elifs or else
```

heevasti: Yes, that is one, only that it looks a bit more messy than I would like to see with respect to the `parse_args` version. But perhaps we can use the `contextlib` to solve this problem, https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager, by decorating the `parse_args`?

Other option which we have used with the services is the https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack, where we then manually call the `enter_context`.

rbudhrani: the `contextlib` method looks nice! If you try it out lemme know if it works

qmi/instruments/bristol/fos.py
```diff
@@ -1,98 +1,248 @@
...
-    global uldaq
-    if uldaq is None:
-        import uldaq  # pylint: disable=W0621
+    global uldaq, ul, enums
+    _logger.debug("Importing %s modules", sys.platform)
+    if (sys.platform.startswith("linux") or sys.platform == "darwin") and uldaq is None:
+        import uldaq  # type: ignore
 
+    elif sys.platform.startswith("win") and (ul is None or enums is None):
+        from mcculw import ul, enums  # type: ignore
```
rbudhrani, line 43: when importing earlier for typing checking `ul` and `enums` are imported but here they are imported from `mcculw`. Should we do the same for type checking?

qmi/instruments/bristol/fos.py
```diff
@@ -1,98 +1,248 @@
...
+        Unix-compatible driver version is instantiated.
+
+        Attributes:
+            fos:        The instantiated FOS device on the driver.
+
+        Parameters:
+            context:    A QMI_Context instance.
+            name:       Name for the instrument in the context.
+            unique_id:  An unique identification number, a.k.a. serial number of the device.
+            board_id:   Board number for Windows driver. Not used for Unix driver.
```
rbudhrani, line 210: if `board_id` is not used for unix, should we make it optional instead of setting a default value of 0. Because 0 does hold meaning when its a Windows FOS, so if not set then a value of 0 is still valid.

heevasti: I think it is simpler this way. We simply do not forward the `board_id` to the Unix driver version, and then we don't need to do any `assert is not None` step before forwarding it to the Windows version.

rbudhrani (approved):

PR #106, opened 2024-10-09
Add rtscts keyword to TransportDescriptorParser

pieterbuzing:
Added a proper case for bool type in the parser (otherwise any non-empty string would evaluate to True).

### DIFF
--- a/qmi/core/transport.py
+++ b/qmi/core/transport.py
@@ -293,6 +293,10 @@ class TransportDescriptorParser:
                     ty = self._keywords[k][0]
                     if ty is int and v.startswith('0x'):
                         parameters[k] = int(v, 16)
+                    elif ty == bool:
+                        if v not in ("True", "False"):
+                            raise ValueError()
+                        parameters[k] = (v == "True")
                     else:
                         parameters[k] = ty(v)
                 except ValueError:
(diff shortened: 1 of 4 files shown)


qmi/core/transport.py
```diff
@@ -293,6 +293,10 @@ def _parse_keyword_parameters(self, strings: List[str]) -> Dict[str, Any]:
                     ty = self._keywords[k][0]
                     if ty is int and v.startswith('0x'):
                         parameters[k] = int(v, 16)
+                    elif ty == bool:
+                        if v not in ("True", "False"):
+                            raise ValueError()
```
heevasti, line 298: The validation of the parameter will be done in the `self._validate...` calls, so this is a double check. Considering that we do need to check the exact string to be correct, it is probably better to do it here, and then in practise we do not need the `validate` method in the `QMI_SerialTransport` class. In principle one _could_ call `QMI_SerialTransport` directly without using our `create_transport`  but this is not instructed in our use of QMI. So, we can drop the second check using the validation method, but it can also stay in case of unintended use. I'm fine with both.

qmi/core/transport.py
```diff
@@ -435,6 +441,11 @@ def _validate_device_name(device: str) -> None:
         if not (device.upper().startswith("COM") or device.startswith("/")):
             raise QMI_TransportDescriptorException("Unknown serial port device path ({})".format(device))
 
+    @staticmethod
+    def _validate_rstcts(rtscts: bool) -> None:
+        if rtscts not in (True, False):
```
heevasti, line 446: This does accept inputs `0` and `1` as well, btw. If you want to make this strict(er), you can add also type check.

heevasti (approved):
Minor comments, but should not have an effect for users as long as drivers with serial connection always are implemented using the transport string and `create_transport`.

PR #112, opened 2024-10-17
[QMI-111] Remove support for Python 3.9 and add support for 3.12.

heevasti:
Removed Support for 3.9 and added support for 3.12.

### DIFF
--- a/.github/workflows/pull-request-ci.yml
+++ b/.github/workflows/pull-request-ci.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       max-parallel: 3
       matrix:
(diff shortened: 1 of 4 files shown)


rbudhrani (changes requested):
Need to edit the following 2 files:
- `TESTING.md`: CI tests are now run with 3.10 and above. There is a reference to 3.9 here
- `documentation/sphinx/source/installation.rst`: There is a line here that says "QMI depends on Python 3.9 or newer." that needs to be updated to 3.10

rbudhrani (approved):

PR #126, opened 2024-12-16
125 new mypy issues to fix

heevasti:
The `analog_discovery` error was fixed by typing `sample_record: np._typing.NDArray[np.float64]` and returning `list(sample_record)` instead of `sample_record.to_list()` where the latter gives a range of possible typings.

The `_picoscope.py` was fixed by typing `buffers: list[ctypes.Array[ctypes.c_int16]]` and `samples: np._typing.NDArray[np.int16]`, and then converting the `buffers[chan]` ctypes arrays into Numpy arrays with `np.ctypeslib.as_array(buffers[chan])`

### DIFF
--- a/qmi/instruments/digilent/analog_discovery.py
+++ b/qmi/instruments/digilent/analog_discovery.py
@@ -383,4 +383,4 @@ class AnalogDiscovery2(QMI_Instrument):
         except pydwf.DwfLibraryError as library_error:
             raise QMI_InstrumentException(str(library_error)) from library_error
 
-        return sample_record.tolist()
+        return list(sample_record)
--- a/qmi/instruments/digilent/analog_discovery.py
+++ b/qmi/instruments/digilent/analog_discovery.py
@@ -352,7 +352,7 @@ class AnalogDiscovery2(QMI_Instrument):
             raise QMI_UsageException("Device should be opened before use.")
 
         analog_in = self._device.analogIn
-        sample_record: np.ndarray = np.array([])
+        sample_record: np._typing.NDArray[np.float64] = np.array([])
         try:
             if amount_samples is None:
                 amount_samples_rational = analog_in.recordLengthGet() // analog_in.frequencyGet()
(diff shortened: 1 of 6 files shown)


qmi/instruments/digilent/analog_discovery.py
```diff
@@ -316,53 +328,59 @@ def prepare_analog_input_record(self,
...
-                unexpected instrument error occurs.
+            QMI_UsageException:      Raised when a device is used before it is opened.
+            QMI_InstrumentException: Raised when data is corrupt or lost or when an unexpected instrument error occurs.
         """
         if self._device is None:
             raise QMI_UsageException("Device should be opened before use.")
 
         analog_in = self._device.analogIn
-        sample_record: np.ndarray = np.array([])
+        sample_record: np._typing.NDArray[np.float64] = np.array([])
```
pieterjbotma, line 355: why are you using `np._typing` instead of just `numpy.typing`: https://numpy.org/devdocs/reference/typing.html#numpy.typing.NDArray

qmi/instruments/picotech/_picoscope.py
```diff
@@ -15,13 +15,13 @@
...
     from picosdk.ps4000a import ps4000a
-    _ps: Union[ps3000a, ps4000a, None] = None
+    _ps: ps3000a | ps4000a | None = None
 
 else:
     _ps = None
 
 # Global dictionary for the imported library commands.
-COMMAND_DICT: Dict = dict()
+COMMAND_DICT: dict[str, Any] = dict()
```
pieterjbotma, line 24: is it really `Any`, can't we constrain it?

qmi/instruments/digilent/analog_discovery.py
heevasti: Oh, my IDE said that there's no reference to 'typing' in '__init__.pyi'.  But it does work if I do remove the underscore. So, I'll change it.

qmi/instruments/picotech/_picoscope.py
heevasti: You're right, it can be a `Callable` I think. I'll try to push with that.

heevasti: You're right, it can be a Callable I think. I'll try to push with that.

pieterjbotma (approved):

PR #127, opened 2024-12-16
124 update highfinesse support files with the license terms

heevasti:
Added the HighFinesse driver, CLI and unit-tests into the codebase. The support files now include the license terms as well.


pieterjbotma (approved):
Looks good!

PR #129, opened 2024-12-18
128 add extra rpc methods for high finesse wlm

heevasti:
Addition of new RPC methods 
- `get_operation_state`
- `get_power`
- `set_data_pattern`
- `get_data_pattern_data`

Refactored typing and some of the existing RPC methods. Refactored Dosctrings. Refactored driver class name and file names. Refactored unit-tests.


qmi/instruments/high_finesse/wlm.py
```diff
@@ -124,35 +186,130 @@ def get_frequency(self, channel: int) -> float:
...
-        if value > 0:
-            return value
+        power = self._lib.dll.GetPowerNum(channel, 0.0)
 
-        err_msg = WS8_ERR(value) if value in [error.value for error in WS8_ERR] else str(value)
+        return self._check_for_error_code(power, "GetPowerNum")
 
-        _logger.warning("[%s] %s error: %s", self._name, method, err_msg)
+    @rpc_method
+    def read_measurement(self, channel: int) -> Measurement:
```
pieterjbotma, line 222: My preference would be to do these type of abstractions outside the drivers. For example, the wavelength monitoring service, we can have a generic/abstract container that we can pipe data from bristol/wlm into. So I think the easiest is just to remove this method and the import to bristol.

pieterjbotma (changes requested):

pieterjbotma (approved):

PR #132, opened 2025-01-07
131 update badge references to version tag of release

heevasti:
Created a release procedure in `RELEASE.md`. It involves three bump2version config files due to restrictions of the regex use (raw strings only) in the configs, and also due to logic working only in one direction (up up up).


RELEASE.md
```diff
@@ -0,0 +1,64 @@
...
+5. Amend the CHANGELOG.md and README.md to the previous release bump commit:
+```shell script
+git add CHANGELOG.md README.md
+git commit --amend
+```
+6. Push these files and see that the pipelines pass.
+```shell script
+git push --set-upstream origin stable-0-46
+```
+7. Tag the release:
```
pieterjbotma, line 31: i don't think we want to bump versions right? Wouldn't it be better to just make two verrsions of the config and then just make a single call? We can then reduce the steps 2-7 to just:
```
2. Replace in `CHANGELOG.md` the `## \[x.y.z] - Unreleased` line with `## [<major>.<minor>.<patch>] - <yyyy>-<mm>-<dd>` line.
3. Amend the CHANGELOG.md and README.md to the previous release bump commit:
```shell script
git add CHANGELOG.md README.md
git commit --amend
4. Run `bump2version` to increment the release:
```shell script
bump2version release --commit --tag --allow-dirty

```

pieterjbotma (changes requested):

RELEASE.md
heevasti: OK, I changed the order of the bump2version commands. The first `bump2version release ...` command now creates also the tag together with the commit. The second bump2version command was too early and now moved to be after making sure the tag was pushed and Pypi package release made. This second command changes the badge lines from `v<major>.<minor>.<patch>` back to `main`, as the third command expects to find `main` string in the badge lines. The third bump2version command bumps to the new beta version, but also bumps now the badge strings as the beta as well. The fourth bump2version command is now needed to change these __again__ back as `main`.

heevasti: New edits on the file to leave the stable branch alive and to make a new branch for updating to new beta version, which is merged to main.

RELEASE.md
```diff
@@ -0,0 +1,97 @@
...
+    ```
+4. Cherry-pick the patch into the stable branch:
+    ```shell script
+    git cherry-pick -x {commit-hash}
+    ```
+5. Restore the `CHANGELOG.md` and update it with the relevent change/fix, and stage it.
+    ```shell script
+    git restore CHANGELOG.md
+    git add CHANGELOG.md
+    ```
```
heevasti, line 86: This I would separate into two steps to avoid doing both `git` commands after another and then realizing the `CHANGELOG.md` needed some manual editing in between.

RELEASE.md
```diff
@@ -0,0 +1,97 @@
...
+    ```shell script
+    bump2version release --config-file=.bumpversion_release.cfg  --allow-dirty --commit --tag
+    ```
+7. Push the branch and tag to origin:
+    ```shell script
+    git push origin stable-{major}-{minor} v{major}.{minor}.{patch} 
+    ```
+8. If the tag passes the workflow, do the following:
+  - Find the tag in GitHub and make a release with the latest `CHANGELOG.md` entry.
+  - See that the release has created a package and upload it to Pypi.
```
heevasti, line 97: upload -> uploaded

RELEASE.md
```diff
@@ -0,0 +1,97 @@
...
+    ```shell script
+    git push --set-upstream origin stable-<major>-<minor>
+    ```
+5. Push the tag to origin:
+    ```shell script
+    git push origin v{major}.{minor}.{patch} 
+    ```
+6. If the tag passes the workflow, do the following:
+  - Find the tag in GitHub and make a release with the latest `CHANGELOG.md` entry.
+  - See that the release has created a package and upload it to Pypi.
```
heevasti, line 33: upload -> uploaded

pieterjbotma (approved):

PR #133, opened 2025-01-07
[QMI-130] Fixing the Mypy check issues as described in the ticket

heevasti:
plus extra changes for new typing style and code style.


pieterjbotma (approved):
Looks good!

PR #160, opened 2025-08-25
159 class docstring wrapper

heevasti:
Added the automated editing of docstrings when creating an RPC proxy of an RPC object (instrument, task...). Also, did lots of pylint score improvements by converting formatted strings as f-strings and updated typing in all `core` files and also in some other files. The important files to review are 

- qmi.core.rpc
- documentation/sphinx/source/tutorial.rst
- tests.core.test_rpc


qmi/core/rpc.py
```diff
@@ -1029,13 +1034,18 @@ def make_interface_descriptor(rpc_object_class: Type[QMI_RpcObject],
         signature = str(inspect.signature(member))
         docstring = member.__doc__
         methods.append(RpcMethodDescriptor(name, signature, docstring))
+        if not name.startswith("_"):
+            signature_doc = signature.replace("(self, ", "(").replace("(self", "(")
```
pieterjbotma, line 1038: is there a reason you have `.replace("(self, ", "(")` twice?

qmi/core/rpc.py
```diff
@@ -1029,13 +1034,18 @@ def make_interface_descriptor(rpc_object_class: Type[QMI_RpcObject],
...
+            doc += f"  - {name}{signature_doc}\n"
 
     # Extract signal declarations.
+    doc += '\nsignals:\n'
```
pieterjbotma, line 1042: minor: use `RPC signals:` to keep it consistent with `RPC methods`

qmi/core/rpc.py
```diff
@@ -1044,16 +1054,19 @@ def make_interface_descriptor(rpc_object_class: Type[QMI_RpcObject],
             constant_names.update(getattr(base, "_rpc_constants"))
 
     # Extract constant values.
+    doc += '\nconstants:\n'
```
pieterjbotma, line 1057: minor: use `RPC constants:` to keep it consistent with `RPC methods:`.

pieterjbotma, line 1057: minor: maybe we can do a check first to see if there are any constants? otherwise (for most RPC objects) we'll just have this blank.

qmi/core/rpc.py
pieterjbotma, line 1042: minor: maybe we can do a check first to see if there are any signals? otherwise (for most RPC objects) we'll just have this blank.

pieterjbotma (dismissed):

PR #174, opened 2025-12-18
[173-support-4x2-and-2x4-mode-in-hdawg] 

heevasti:
Refactoring of the HDAWG driver to allow other AWG core <-> channel grouping modes. Also unit-tests fixed.

This PR is still using the same "old" zhinst packages as before, and not the new "toolkit" packages. The ticket turned out to be quite large when doing together with the refactoring to use the toolkit packages, so it is now split into two.

A couple of TODO:s are remaining related to the "open" and "close" as it is not clear to me if only checking the state of the currently selected AWG core index in the `awg_module` is enough when the instrument is in another grouping mode than 1x8. If you have some comments on that, I would like to hear it.

### DIFF
--- a/qmi/instruments/zurich_instruments/hdawg.py
+++ b/qmi/instruments/zurich_instruments/hdawg.py
@@ -18,6 +18,7 @@ import numpy as np
 
 # Lazy import of the zhinst module. See the function _import_modules() below.
 if TYPE_CHECKING:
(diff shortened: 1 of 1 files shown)


heevasti:
In the end we include also the update to use zhinst-toolkit into one pull request (#181). Closing this pull request.

PR #175, opened 2026-01-05
135 make pololu maestro driver be possible to use in uart mode as well

### LINKED ISSUE #135: Make Pololu Maestro driver be possible to use in UART mode as well.
The QMI Pololu Maestro device driver is now supporting the use of the device in Dual USB or Daisy Chain modes. These are not the default modes, but an UART mode is the default of the device. So, to use the driver, the manufacturer-provided GUI needs to be used first to change the mode.

First of all, this should be mentioned in our driver. The documentation should be edited to instruct the mode change for new devices. Secondly, we could enable the UART mode as well by selectively importing a separate class for UART control, or to make the driver smart enough to recognize or figure out which mode the device is on.

The original driver that was developed for Diamondos was based largely on the example of [Maestro](https://github.com/FRC4564/Maestro/tree/master) package's `maestro.py`. The `umaestro.py` is the UART version and looks very similar in function to the `maestro.py`.

The UART control is rather similar to the serial-to-USB control through COM ports. We could create a special transport string that recognizes if we want to use the UART mode or not. Something like
```python
if transport.upper().startswith("UART"):
    self._transport = busio.UART(board.TX, board.RX, baudrate)
```
From the first look the existing `_write` and `_ask` functions could be used for both control ways without any or little modification. Probably the stumbling point here is the `self._transport.discard_read()`

**Acceptance criteria**
Please specify the acceptance criteria in as much detail as possible,
for example:

- if this feature is available how would you invoke it (code example)
- what would the output look like?
    - if it is an object, what would the members be?
    - if it is a plot, a sketch would help
    - anything else that clarifies what would constitute a satisfactory
      implementation?

**Additional context**

heevasti:
Pololu Maestro QMI driver was updated so that it can be used also with Pololu instruments that are in the default UART mode, by using a special "uart:..." transport string. For this we need `board` and `busio` modules from `adafruit-blinka` package, which are imported as "lazy imports". The latter is imported through new `qmi_uart` module, which has an extension of the `busio.UART` class so that it is compatible with `QMI_Transport` classes to use with the special transport string, and to use the same read-write calls.

During the course of this refactoring and extension, also the `psutil` package was updated such that it would not work anymore in `qmi_proc.py`. This is due to moving the `_snicaddr` internal variable in the package. The `qmi_proc.py` was adapted such that it checks the version number of  `psutil` and selects followingly the correct location of the `_snicaddr`.

### DIFF
--- a/qmi/instruments/pololu/maestro.py
+++ b/qmi/instruments/pololu/maestro.py
@@ -13,7 +13,7 @@ import warnings
 
 # Lazy import of the adafruit-blinka module 'board' and 'QMI_Uart'. See the function _import_modules() below.
 if TYPE_CHECKING:
-    import board
+    import board  # mypy: ignore
     from qmi.instruments.pololu.qmi_uart import QMI_Uart
 else:
     board = None
--- a/qmi/instruments/pololu/maestro.py
+++ b/qmi/instruments/pololu/maestro.py
@@ -13,7 +13,7 @@ import warnings
 
 # Lazy import of the adafruit-blinka module 'board' and 'QMI_Uart'. See the function _import_modules() below.
 if TYPE_CHECKING:
-    import board  # mypy: ignore
+    import board  # type: ignore
     from qmi.instruments.pololu.qmi_uart import QMI_Uart
 else:
     board = None
--- a/qmi/instruments/pololu/maestro.py
+++ b/qmi/instruments/pololu/maestro.py
@@ -1,38 +1,71 @@
 """Instrument driver for the Pololu maestro servo controller"""
 
 import logging
-from typing import Dict, Generator, List, Optional, Tuple
+from typing import TYPE_CHECKING, Generator
 from time import sleep
 
 from qmi.core.context import QMI_Context
 from qmi.core.instrument import QMI_Instrument, QMI_InstrumentIdentification
 from qmi.core.rpc import rpc_method
-from qmi.core.transport import create_transport
+from qmi.core.transport import create_transport, SerialTransportDescriptorParser, QMI_Transport
 from qmi.core.exceptions import QMI_InstrumentException
 import warnings
+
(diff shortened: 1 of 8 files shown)


qmi/instruments/pololu/maestro.py
```diff
@@ -1,38 +1,72 @@
...
+    board = None
+    QMI_Uart = None
+
 # Global variable holding the logger for this module.
 _logger = logging.getLogger(__name__)
 
 
-class Pololu_Maestro(QMI_Instrument):
-    """Instrument driver for the Pololu Maestro servo controller."""
+def _import_modules() -> None:
```
bvanommen, line 26: maybe change the name so it's more clear it's about the uart support?

qmi/instruments/pololu/maestro.py
```diff
@@ -287,11 +331,11 @@ def set_target(self, channel: int, value: int) -> None:
         self._targets[channel] = target
 
     @rpc_method
-    def get_targets(self) -> List[int]:
+    def get_targets(self) -> list[int]:
         """Get the current targets for each channel.
 
         Returns:
-            The target values for each channel.
+            self_targets: The target values for each channel.
```
bvanommen: self._targets (typo?)

qmi/instruments/pololu/maestro.py
```diff
@@ -394,11 +445,15 @@ def set_max_target(self, channel: int, value: int) -> None:
         Parameters:
             channel:    The channel number.
             value:      The new maximum target value.
+
+        Raises:
+            ValueError: If given maximum target value is not in valid range.
         """
         if value <= self._min_target_values[channel] or value > self.DEFAULT_MAX_VALUE:
```
bvanommen: here it's <= but in the previous set_min_target function it's > <. Is there a reason for the difference? I would advocate in general to not be able to set min=max

PR #198, opened 2026-04-09
197: Make `Context` objects context managers

Thom747:
This PR implements #197.


qmi/core/context.py
```diff
@@ -11,28 +12,26 @@
 import string
 import threading
 import time
+import typing
```
heevasti: This is used to return `typing.Literal` at the `__exit__`. But we should import `Literal` below with the other `typing`  imports directly.

qmi/core/context.py
```diff
@@ -290,6 +289,42 @@ def __init__(self, name: str, config: CfgQmi | None = None) -> None:
...
+            # If a context is acquired through `qmi.start()`, it will already have been started, so we skip starting.
+            self.start()
+        return self
+
+    def __exit__(
+            self,
+            exc_type: type[BaseException] | None,
+            exc_val: BaseException | None,
+            exc_tb: TracebackType | None,
+    ) -> typing.Literal[False]:
```
heevasti: Simplify, with the direct import of `Literal` to `) -> Literal[False]:`

qmi/core/context.py
```diff
@@ -290,6 +289,42 @@ def __init__(self, name: str, config: CfgQmi | None = None) -> None:
...
+        """
+        Context manager exit point.
+        Stops the context if it was not already stopped.
+
+        Arguments:
```
heevasti: I have typically used `Parameters` instead of `Arguments` throughout QMI. Its a minor point but for uniformity it would be nice to use the same nomenclature.

qmi/core/context.py
```diff
@@ -290,6 +289,42 @@ def __init__(self, name: str, config: CfgQmi | None = None) -> None:
...
+            exc_type: Type of the exception that caused the context to be exited.
```
heevasti, line 319: Another minor formatting detail: I tend to align the text after the colon to start from the same column as the longest parameter name starts from. So here I would change the text to: 
```
            exc_type: Type of the exception that caused the context to be exited.
            exc_val:  Exception that caused the context to be exited.
            exc_tb:   Traceback of the exception that caused the context to be exited.
```

qmi/core/context_singleton.py
```diff
@@ -107,6 +104,9 @@ def start(
         init_logging:     Optional flag; set False to skip logging initialization.
         console_loglevel: Optionally override `console_loglevel` from config file.
         context_cfg:      Optionally insert or override context(s) in config.contexts.
+
+    Returns:
+        The global QMI_Context.
```
heevasti: Could change this to:
`_qmi_context: The global QMI_Context object.`

CHANGELOG.md
```diff
@@ -23,6 +23,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - DeprecationWarning on the `set|get_output_amplitude` methods in the HDAWG driver, as the new firmware points to rather using `outputs/n/gains/n`. Use `set|get_output_gain` from now on.
 - All modules in `qmi.data` were made compatible also with `h5netcdf` package. It can now be used equivalently with the `h5py`-based HDF5 data files when providing `backend="h5netcdf"` input parameter on specific class initializations and calls.
 - Changed Thorlabs K10CR1 to derive from new base class for K10CRx instruments.
+- DeprecationWarning warning on the `start_stop` context manager, as `Context` objects are now their own context managers.
```
heevasti: These should go under a new section `### Changed` under the `## [0.53.0-beta.0] - Unreleased`.

heevasti (changes requested):
Great work, nice to see it works so nicely. I also tested it with the Yokogawa I have here on the desk now and it worked without issues. A few comments, mainly related just to imports and formatting. And CHANGELOG  comments.

qmi/core/context.py
Thom747: Good point, changed!

qmi/core/context.py
Thom747: Done!

qmi/core/context.py
Thom747: Good point. This was wrong in a couple of other places as well, so I have replaced those as well.

qmi/core/context.py
Thom747: Done!

qmi/core/context_singleton.py
Thom747: I don't see why a user of this function would care about the name of the object we return?
I have added the "object" part though.

CHANGELOG.md
Thom747: Ah yeah, my bad, fixed!

qmi/core/context.py
heevasti: Let's limit the changes only to files within this scope.

This because I do also use `Arguments` for class arguments. And let's keep those unchanged at least for now.

So please revert the changes in other modules.

Thom747: Reverted in other modules

heevasti (approved):
Ready to merge.

PR #200, opened 2026-04-10
184 add new functions to yokogawa driver

### LINKED ISSUE #184: Add new functions to Yokogawa driver
The QMI driver for Yokogawa does not seem to be able to obtain trace data, only copying files after a measurement. And event that requires the Yokogawa be mounted as a flash drive through USB. It would be good to add an option that also makes it possible to obtain trace data during a measurement through the ethernet connection.

The following new RPC method was tested to return requested trace data in ASCII format:
```python
    @rpc_method
    def send_waveform(self, channels: list[int]):
        """
        Returns the onscreen traces from specified channels in the form on a 2D numpy array. Data is streamed via ethernet
        in the 'ascii' format.

        Parameters:
            channels:           A list of the channels to return data from
            data_type:          What format to send the data in

        Returns:
            traces:             An array where each row is a channel trace. Channel order correspond to 'channels' parameter 
        """
        # Expected length of each trace
        num_points = self.get_number_data_points()

        # Initialize data array
        traces = np.zeros((len(channels), num_points))

        # Stop to acquire data
        self._scpi_protocol.write(":STOP")

        # Set data type (could also be binary, but then must adjust how 'resp' is processed below)
        self._scpi_protocol.write(":WAVeform:FORMat ASCII")

        # For each channel
        for i, val in enumerate(channels):
            # Set which channel to get data from
            self._scpi_protocol.write(f":WAVeform:TRACE {val}")
            # Transfer data from the stopped acquisition
            resp = self._scpi_protocol.ask(":WAVeform:SEND?")
            traces[i] = np.array([float(x) for x in resp.split(',')])

        # Start continuous acquisition again
        self._scpi_protocol.write(":START")
        return traces
```
So we want to implement a call that returns trace data for each channel defined in the input.

It would be preferrable, though, that the above function gets chopped into smaller pieces. Also, for the completeness, it would be nice that the waveforms could be obtained (and converted to Volts) in all data formats (not data _types_, those are for the files). 

It sh|could be refined to:
a) Have a more descriptive name like `get_waveforms_data` or `get_waveforms` or `get_trace_data` or such.
b) Create separate calls for setting the data format (`set_data_format`) and setting the waveform trace channel (`set_waveform_trace_channel`), which then would be called within the above call.
c) Setting the data format is probably not necessary to be in the call, as the format will be permanent after being set once, unless someone | something changes it. It should suffice in a script to call this once before calling to obtain data.
d) Consulting the manual, and testing on the hardware, the data streaming should be enabled in all the available data formats. For that, we should also create a separate method that obtains the trace data and manipulates it according to the data format (`get_waveform_trace_data`).

The last one also raises the need of extra calls to obtain waveform data range, offset and position, which are needed for BYTE, RBYTE and WORD data format conversions to Volts.

**Acceptance criteria**
The updated driver works also on hardware, and can acquire data on all data formats (correctly).

**Additional context**
N/A

heevasti:
Addition of new data tracing functions with optional data formats.

Edited also `ScpiProtocol.ask` to allow to set another decoder, as the other data formats than "ASCII" set the Yokogawa device return byte strings that are "latin1" encoded.


qmi/core/scpi_protocol.py
```diff
@@ -62,16 +63,19 @@ def write_raw(self, cmd: bytes) -> None:
         binary_cmd = cmd + self._command_terminator
         self._transport.write(binary_cmd)
 
-    def ask(self, cmd: str, timeout: float | None = None, discard: bool = False) -> str:
+    def ask(self, cmd: str, timeout: float | None = None, discard: bool = False, decoder: str = "ascii") -> str:
```
Thom747, line 66: Might it make sense to use a `StrEnum` to specify a set of known encodings?
Currently, it is unclear to me from this annotation or the documentation what valid values would be for this argument.

qmi/core/scpi_protocol.py
```diff
@@ -62,16 +63,19 @@ def write_raw(self, cmd: bytes) -> None:
...
+            cmd:     SCPI command string.
             timeout: Optional response timeout in seconds.
             discard: Discard contents in read buffer before asking. Default is False.
+            decoder: Optional parameter to set another decoder to use for the query. Some instruments
+                     have optional data format settings which can then return data in other than 'ascii'
+                     coded format.
 
         Returns:
-            Response message with message terminator removed.
+            decoded_response: Decoded response message with message terminator removed.
```
Thom747, line 78:
```suggestion
            Decoded response message with message terminator removed.
```

qmi/core/scpi_protocol.py
```diff
@@ -85,26 +89,32 @@ def ask(self, cmd: str, timeout: float | None = None, discard: bool = False) ->
...
         # Read response.
         response = self._transport.read_until(message_terminator=self._response_terminator, timeout=timeout)
-        if not response.endswith(self._response_terminator):
+        if decoder.lower() != "ascii":
+            _response_terminator = self._response_terminator.decode("ascii").encode(decoder)
+        else:
+            _response_terminator = self._response_terminator
+
+        if not response.endswith(_response_terminator):
             raise QMI_InstrumentException("Bad response")
```
Thom747:
```suggestion
            raise QMI_InstrumentException("Response was not terminated with expected terminator.")
```

qmi/instruments/yokogawa/dlm4038.py
```diff
@@ -61,6 +61,20 @@ def _channels_check(self, channels: int | list[int] | str) -> int | list[int]:
...
+    def _data_format_check(data_format: str) -> str:
+        """Check the data format string and return in SCPI style."""
+        data_formats = {
+            "ascii": "ASCii",
+            "byte": "BYTE",
+            "rbyte": "RBYTe",
+            "word": "WORD"
+        }
+        if data_format.lower() not in data_formats:
+            raise QMI_InstrumentException(f"Unexpected data format, got {data_format}")
```
Thom747: Minor: the pythonic way of doing things is to "ask for forgiveness, not permission". You would thus optimistically get the value and then raise `QMI_InstrumentException` from the resulting `KeyError`.

qmi/instruments/yokogawa/dlm4038.py
```diff
@@ -325,49 +355,183 @@ def set_number_data_points(self, length: int) -> None:
...
+        """Gets the number of data points that will be saved."""
+        return int(self._scpi_protocol.ask(":WAVeform:LENGth?")[10:])
+    
+    @rpc_method
+    def set_data_format(self, data_format: str) -> None:
+        """Set the waveform data format to specific type.
+
+        Parameters:
+            data_format: In which format to send the data in. Possible options:
+                         'ASCii', 'BYTE', RBYTE', 'WORD'.
```
Thom747, line 374: This feels like a natural fit for a `StrEnum`.

qmi/instruments/yokogawa/dlm4038.py
```diff
@@ -325,49 +355,183 @@ def set_number_data_points(self, length: int) -> None:
...
-             average: Has to be between 2 and 1024 in 2**n steps.
+    @rpc_method
+    def get_waveform_trace_data(self, data_format: str) -> np.ndarray:
+        """Get waveform (channel) trace data from single waveform.
+
+        In ASCII format the data is in a comma-separated list of float voltage values.
+
+        In any other format, the data is as block data. Block data is signed 8-bit data in 'BYTE' and 'RBYTe'
+        data formats. 'RBYTe' has also reverse order of data points. In 'WORD' format the data block contains
+        signed 16-bit data in little-endiand order. The block data syntax is as follows:
```
Thom747:
```suggestion
        signed 16-bit data in little-endian order. The block data syntax is as follows:
```

qmi/instruments/yokogawa/dlm4038.py
```diff
@@ -325,49 +355,183 @@ def set_number_data_points(self, length: int) -> None:
...
+        "#N<N-digit decimal number><data byte sequence>"
+
+        The block data needs to be converted into Volts. The respective conversion formula depends on the
+        data format. See the DLM4000 series user's manual for details.
```
Thom747, line 413: Is it possible to perform this conversion for the user for all encodings?

qmi/instruments/yokogawa/dlm4038.py
```diff
@@ -325,49 +355,183 @@ def set_number_data_points(self, length: int) -> None:
...
+            return np.array([float(x) for x in raw_data.split(',')])
+
+        # Check that we have block data format, where the string must start with a hash.
+        if not raw_data.startswith('#'):
+            raise QMI_InstrumentException(f"Unexpected response {raw_data!r} for ':WAVeform:SEND?'.")
+        
+        # 'N' indicates the number of succeeding data bytes (digits) in ASCII code.
+        no_of_bytes = int(raw_data[1])
+        # <N-digit decimal number> indicates the number of bytes of data.
+        data_points = int(raw_data[2:2+no_of_bytes])
```
Thom747, line 441: How I read this, you are taking a potentially very large array of bytes and converting it to an `int`. I have never done such a conversion before, but I know Python supports arbitrary length integers, so I assume this results in a single `int`. Does that `int` mean anything though? To me, it does not seem to mean "the number of bytes of data" that the comment says this variable should contain. Could you explain?

qmi/instruments/yokogawa/dlm4038.py
```diff
@@ -325,49 +355,183 @@ def set_number_data_points(self, length: int) -> None:
...
+                         Channel order corresponds to 'channels' input parameter.
+        """
+        # Expected length of each trace
+        num_points = self.get_number_data_points()
+
+        # Initialize data array
+        traces = np.zeros((len(channels), num_points))
+
+        # Stop to acquire data
+        self.stop()
```
Thom747, line 503: Should we use a context manager to ensure we start again later on?

Thom747, line 503: Apart from the context manager, I was also wondering if it is a problem that we are effectively stopping data collection here. Would that not possibly result in data loss?

CHANGELOG.md
```diff
@@ -23,7 +30,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - DeprecationWarning on the `set|get_output_amplitude` methods in the HDAWG driver, as the new firmware points to rather using `outputs/n/gains/n`. Use `set|get_output_gain` from now on.
 - All modules in `qmi.data` were made compatible also with `h5netcdf` package. It can now be used equivalently with the `h5py`-based HDF5 data files when providing `backend="h5netcdf"` input parameter on specific class initializations and calls.
 - Changed Thorlabs K10CR1 to derive from new base class for K10CRx instruments.
-- Replace `pylint` linter with `ruff`.
```
Thom747: My bad, thanks for the fix!

Thom747 (changes requested):
Overall, this looks good! I mostly have some minor remarks and questions.

qmi/core/scpi_protocol.py
heevasti: There are [so many](https://docs.python.org/3/library/codecs.html#standard-encodings) that I would not make an enum over them. Just let it except if the user uses an invalid value.

qmi/core/scpi_protocol.py
heevasti: Changed to this.

qmi/core/scpi_protocol.py
heevasti: I think it is fine with the returned variable name as well.

qmi/instruments/yokogawa/dlm4038.py
heevasti: Wouldn't that need a `try-except`? Seems more complicated than this simple check. Perhaps it is faster but I suppose we do not have to worry about speed here.

qmi/instruments/yokogawa/dlm4038.py
heevasti: OK, I will use `StrEnum` for these.

qmi/instruments/yokogawa/dlm4038.py
heevasti: Fixing.

qmi/instruments/yokogawa/dlm4038.py
heevasti: ASCII gives Volts directly, and for the other possible encodings, 'WORD', 'BYTE' and 'RBYTE', the conversion is done at the return statement line. So I would say yes. The currently not-implemented optional parameter `<NRf>` just selects the return type of values be integer, fixed-point or floating-point number.

qmi/instruments/yokogawa/dlm4038.py
heevasti: It means the number of data bytes in the returned "raw_data" string after position 2+no_of_bytes. So, "#6000123..." would indicate that there are six bytes that indicate how many data bytes would follow (in the '...' part). So, in this example '...' would be 123 bytes long.

qmi/instruments/yokogawa/dlm4038.py
heevasti: It is hard-coded in the method so creating a context manager for this would be an overkill.

heevasti: The manual says you have to stop the data acquisition to be able to obtain the trace data. I tried it actually to obtain without stopping, but then you don't get any data. The user just needs to know that while getting trace data there is no acquisition. Hardware limitation that we cannot change.

qmi/core/scpi_protocol.py
Thom747: I see. Would it make sense to link to that page in the documentation, so that users know that these are the encodings we are talking about?

qmi/instruments/yokogawa/dlm4038.py
Thom747: Ah, I see. Good!

qmi/instruments/yokogawa/dlm4038.py
Thom747: Ah, too bad

Thom747 (dismissed):

Thom747 (approved):

PR #205, opened 2026-04-15
204: Use no configuration file if the `qmi.start` parameter `config_file` is explicitly set to `None`

### LINKED ISSUE #204: Implement "None means None" config_file argument to `qmi.start()`
Description 
=========== 
Calls to `qmi.start()` take in a second parameter called `config_file` with type `str | None`.
Currently, if the user specifies an explicit `None` argument, QMI will look up the `QMI_CONFIG` environment variable and use that instead of following the stated intention of using no predefined configuration file (an "empty configuration).
This is counter-intuitive and can lead to unexpected behaviour.
Thus, the aim of this issue is to replace this behaviour with a situation where explicitly specifying `None` results in no configuration being used.

Modules to be created 
--------------------- 
None

Modules to be modified
---------------------- 
`qmi/core/context_singleton.py`.

Tests to be created/updated 
--------------------------- 
No tests use explicit `None`s.

Documentation to be updated 
--------------------------- 
`CHANGELOG.md`

Hardware 
-------- 
None

Thom747:
Implements #204.


heevasti (changes requested):
Looks fine. But I think there should be a unit-test[s] that would notice that the method has slightly changed, giving `CfgQmi` object back *always* when `None` is explicitly given. And all the other options should work as before.

Thom747:
> Looks fine. But I think there should be a unit-test[s] that would notice that the method has slightly changed, giving `CfgQmi` object back _always_ when `None` is explicitly given. And all the other options should work as before.

Added a unittest!

heevasti (dismissed):
ok, now it is ready.

heevasti (approved):

PR #207, opened 2026-04-15
206: Make `discover_peer_contexts` return `NamedTuple`s to increase usability

### LINKED ISSUE #206: Return structured datatype from `discover_peer_contexts()`
Currently, `Context.discover_peer_contexts()` returns a `list[tuple[str,str]]`.
This means that users of this method need to know the meaning of the fields in the tuple to be able to use it.
 
This method should return a list of some structured public dataclass so that users can know from the method signature what is returned and what its fields are.

This method could also return a list of `NamedTuple`s, but `NamedTuple`s are generally just inferior `dataclass`es.
Using a `NamedTuple` would have the advantage of not breaking backwards compatibility though, as you can still index into them.

**Acceptance criteria**
The `dataclass` returned should have the following fields:
- `context_name:str`
- `address:str`
- `port:int`

Alternatively, address and port may be merged into a single `address_port:str` field.

**Additional context**
See also this discussion: https://github.com/QuTech-Delft/QMI/pull/191#discussion_r3045493187.

Thom747:
This implements #206.

I decided to go the `NamedTuple` route instead of using a `dataclass`, as the `dataclass` approach broke backwards-compatibility.


qmi/core/context.py
```diff
@@ -689,11 +689,18 @@ def has_peer_context(self, peer_context_name: str) -> bool:
         context_names = self._message_router.get_peer_context_names()
         return peer_context_name in context_names
 
+    class PeerDescriptor(NamedTuple):
```
heevasti, line 692: Can we change the class docstring in the style that is used in some other places (not always, though) with `NamedTuples`? For example:
```python
class SignalDescription(NamedTuple):
    """Description of signal type, specifying its name and arguments.

    Attributes:
        name: Signal name.
        arg_types: Tuple of Python types (e.g. `int` or `str`), describing
            the parameters that will be passed in each published instance of
            this signal type.
    """
    name: str
    arg_types: tuple[Type, ...]
```

Thom747: Done!

heevasti (approved):
Ready for merging.

Issue #213, opened 2026-05-21
Changing RPC constants on proxy does not change the constants on the real object


heevasti:
Description
========

The `_rpc_constants` class variable can be used in all `QMI_RpcObjects` to expose the class variables of the real object to its proxy object. This also means that the class variable value can be changed by doing `TheProxyObjectToRealObject.variable_x = new_value`, if it was defined as `_rpc_constants = ["variable_x"]` before the `__init__` method in the real object (let's define it as `RealObject(QMI_RpcObject)`. 

The idea of the "_rpc_constants" is to stay "constant" by name, but at times we would like to change this "constant". For example when we notice some timeout becomes too short for some reason and we want to extend it without having to modify the real object itself. If we only use the proxy of the real object for the new value, it is fine. But we often need the real object's value in internal calls of the real object. 

At the moment QMI lacks the forwarding of constant value change to the real object from the proxy. We need to make request and reply message classes, sending and handling of constant RPC requests in `QMI_RpcFuture`, and pushing and handling them in `_RpcThread`, and handling it in `RpcObjectManager`. These should be done similarly to the respective RPC method call requests and also lock requests.

Best approach is probably to create "constant" versions of `QMI_XxxRpcRequest|ReplyMessage` classes and see where they have been used for methods and locking, and add "constant" version of the use cases there.

Affected components
==============
`qmi.core`

Files to be changed
=============
- `qmi/core/rpc.py`
- `CHANGELOG.md`
- Documentation about RPC calls -> add that constants can be changed too.
- Unit-tests

PR #214, opened 2026-05-21
Changing rpc constants on proxy

heevasti:
Related to #213.

With this, the "RPC constants" that are class constants on a RPC object, e.g. QMI_Instrument, have been renamed to "RPC properties, and can now be modified so that the modifications propagate back to the actual object, and not proxy only. With this, the proxy class of the object was also made prohibitive on adding new class attributes, RPC properties can only be modified with same type (and size if applicable) values, and also class constants not defined with `_rpc_properties` are not modifiable and will lead to an `AttributeError`.

Further checks are made that the given `_rpc_properties` values are not referring to e.g. a method, QMI signal attribute, property, etc. Invalid naming will raise a `QMI_UsageError`.

`_rpc_constants` use will now give a deprecation warning, and will be handled the same way as the `_rpc_properties`.

Some tutorial examples were also added to illustrate this new functionality.

### DIFF
--- a/qmi/core/rpc.py
+++ b/qmi/core/rpc.py
@@ -1340,7 +1340,7 @@ class _RpcThread(QMI_Thread):
         if self._locking_token is None or self._locking_token == request.lock_token:
             # Invoke the method; this can raise an exception or return a result.
             try:
-                method = self._check_and_get_method(request)
+                method: Callable = self._check_and_get_method(request)
                 result_type = QMI_RpcFutureState.RESULT_IS_VALUE
                 result = method(*request.method_args, **request.method_kwargs)
 
@@ -1363,7 +1363,7 @@ class _RpcThread(QMI_Thread):
         )
         return reply
 
-    def _check_and_get_method(self, request: QMI_MethodRpcRequestMessage) -> _T:
+    def _check_and_get_method(self, request: QMI_MethodRpcRequestMessage) -> Callable:
         """Check if the object has the method requested and is RPC callable; if so, return it."""
         assert self._rpc_object is not None
 
@@ -1519,7 +1519,10 @@ class _RpcThread(QMI_Thread):
 
         _logger.debug("Stopping RPC thread")
 
-    def push_rpc_request(self, rpc_request: QMI_MethodRpcRequestMessage | QMI_LockRpcRequestMessage  | None) -> None:
+    def push_rpc_request(
+        self,
+        rpc_request: QMI_MethodRpcRequestMessage | QMI_ConstantRpcRequestMessage | QMI_LockRpcRequestMessage | None
+    ) -> None:
         """Push an RPC request into the request queue and notify the thread."""
         with self._cv:
             self._fifo.append(rpc_request)
--- a/qmi/core/rpc.py
+++ b/qmi/core/rpc.py
@@ -517,8 +517,10 @@ class QMI_RpcFuture(QMI_MessageHandler):
             self._context.send_message(request)
 
         except QMI_MessageDeliveryException as exc:
-            self._set_result(QMI_RpcFutureState.RESULT_IS_EXCEPTION, (exc,
-            traceback.extract_tb(exc.__traceback__)))
+            self._set_result(
+                QMI_RpcFutureState.RESULT_IS_EXCEPTION, (exc,
+                tuple(traceback.format_list(traceback.extract_tb(exc.__traceback__)))
+            ))
 
     def send_lock_rpc_request_message(self, action: QMI_LockRpcAction) -> None:
         request = QMI_LockRpcRequestMessage(self.address, self.rpc_object_address, self.lock_token, action)
@@ -526,8 +528,10 @@ class QMI_RpcFuture(QMI_MessageHandler):
             self._context.send_message(request)
 
         except QMI_MessageDeliveryException as exc:
-            self._set_result(QMI_RpcFutureState.RESULT_IS_EXCEPTION, (exc,
-            traceback.extract_tb(exc.__traceback__)))
+            self._set_result(
+                QMI_RpcFutureState.RESULT_IS_EXCEPTION, (exc,
+                tuple(traceback.format_list(traceback.extract_tb(exc.__traceback__)))
+            ))
 
     def handle_message(self, message: QMI_Message) -> None:
         """Called when a reply message is received."""
@@ -545,9 +549,9 @@ class QMI_RpcFuture(QMI_MessageHandler):
             # Delivery of RPC request failed.
             exc = QMI_MessageDeliveryException(message.error_msg)
             self._set_result(
-                QMI_RpcFutureState.RESULT_IS_EXCEPTION,
-                (exc, traceback.extract_tb(exc.__traceback__))
-            )
+                QMI_RpcFutureState.RESULT_IS_EXCEPTION, (exc,
+                tuple(traceback.format_list(traceback.extract_tb(exc.__traceback__)))
+            ))
         else:
             _logger.error(
                 "Future for %s.%s received unexpected message type %r",
@@ -611,17 +615,16 @@ class QMI_RpcFuture(QMI_MessageHandler):
                             if not isinstance(self._result, BaseException):
                                 raise QMI_RuntimeException("Received invalid exception value from RPC call.")
                             raise self._result
+
                         exc = self._result[0]
-                        tb = self._result[1]
+                        tb_str_tuple = self._result[1]
                         if not isinstance(exc, BaseException):
                             raise QMI_RuntimeException("Received invalid exception value from RPC call.") from exc
 
-                        tb_str_list = traceback.format_list(tb)
                         _logger.error("Error occurred during RPC call. Traceback from RPC method:\n")
-                        _logger.error("".join(tb_str_list))
+                        _logger.error("".join(tb_str_tuple))
                         _logger.error(exc)
                         raise exc
-
 
                     if self._state == QMI_RpcFutureState.RESULT_IS_VALUE:
                         return self._result
@@ -1459,7 +1462,7 @@ class _RpcThread(QMI_Thread):
                 _logger.debug("RPC method call failed", exc_info=True)
                 result_type = QMI_RpcFutureState.RESULT_IS_EXCEPTION
                 # return the exception, and the traceback
-                result = (exception, traceback.extract_tb(exception.__traceback__))
+                result = (exception, tuple(traceback.format_list(traceback.extract_tb(exception.__traceback__))))
         else:
             _logger.error("%s locked, method request without lock token is denied.", self._rpc_object._name)
             result_type = QMI_RpcFutureState.OBJECT_IS_LOCKED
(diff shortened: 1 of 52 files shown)


qmi/core/rpc.py
```diff
@@ -430,55 +526,101 @@ def send_method_rpc_request_message(self,
...
+            ))
 
     def handle_message(self, message: QMI_Message) -> None:
         """Called when a reply message is received."""
 
         if isinstance(message, QMI_MethodRpcReplyMessage):
-            # Received result from RPC call.
+            # Received result from RPC method call.
+            self._set_result(message.state, message.result)
+        elif isinstance(message, QMI_PropertyRpcReplyMessage):
```
bvanommen, line 584: use or?

qmi/core/rpc.py
```diff
@@ -430,55 +526,101 @@ def send_method_rpc_request_message(self,
...
         elif isinstance(message, QMI_ErrorReplyMessage):
             # Delivery of RPC request failed.
             exc = QMI_MessageDeliveryException(message.error_msg)
-            self._set_result(QMI_RpcFutureState.RESULT_IS_EXCEPTION,
-                             (exc, traceback.extract_tb(exc.__trackback__))
-                             )
+            self._set_result(
+                QMI_RpcFutureState.RESULT_IS_EXCEPTION, (exc,
+                tuple(traceback.format_list(traceback.extract_tb(exc.__traceback__)))
+            ))
```
bvanommen, line 596: I have seen in the lab that this error providing may create incompatibility with servers using older QMI versions. Do we care?

qmi/core/rpc.py
```diff
@@ -694,9 +852,49 @@ def make_rpc_forward_function(method_name: str):
...
+            if name in rpc_property_names:
+                rpc_property_call(
+                    object.__getattribute__(self, "_context"),
+                    object.__getattribute__(self, "_rpc_object_address"),
+                    name,
+                    object.__getattribute__(self, "_lock_token"),
+                    value,
+                    set_property_value=True
+                )
+                return
```
bvanommen, line 890: how does the caller know whether the call was a succes or failure?

qmi/core/rpc.py
```diff
@@ -430,55 +526,101 @@ def send_method_rpc_request_message(self,
 
         try:
             self._context.send_message(request)
+
+        except QMI_MessageDeliveryException as exc:
+            self._set_result(QMI_RpcFutureState.RESULT_IS_EXCEPTION, exc)
```
bvanommen: I think I understand, but to double check: we don't expect a tuple of exc and traceback here because send_message just raises message delivery exceptions?

qmi/core/rpc.py
```diff
@@ -430,55 +526,101 @@ def send_method_rpc_request_message(self,
...
+
+        try:
+            self._context.send_message(request)
+
         except QMI_MessageDeliveryException as exc:
-            self._set_result(QMI_RpcFutureState.RESULT_IS_EXCEPTION, (exc, 
-            traceback.extract_tb(exc.__trackback__)))
+            self._set_result(
+                QMI_RpcFutureState.RESULT_IS_EXCEPTION, (exc,
+                tuple(traceback.format_list(traceback.extract_tb(exc.__traceback__)))
```
bvanommen, line 564: But then here we do expect a tuple. I guess we shouldn't do so here either

qmi/core/rpc.py
```diff
@@ -430,55 +526,101 @@ def send_method_rpc_request_message(self,
...
         request = QMI_LockRpcRequestMessage(self.address, self.rpc_object_address, self.lock_token, action)
         try:
             self._context.send_message(request)
+
         except QMI_MessageDeliveryException as exc:
-            self._set_result(QMI_RpcFutureState.RESULT_IS_EXCEPTION, (exc, 
-            traceback.extract_tb(exc.__trackback__)))
+            self._set_result(
+                QMI_RpcFutureState.RESULT_IS_EXCEPTION, (exc,
+                tuple(traceback.format_list(traceback.extract_tb(exc.__traceback__)))
```
bvanommen, line 575: why did we make this change to the returned traceback?

bvanommen: previously we formatted the tb where it gets printed. I guess either is fine?

qmi/core/rpc.py
bvanommen: Now I'm not sure anymore. Doesn't this exception still end up in the wait() function? I guess there it gets handled under the "older QMI version" code path

qmi/core/rpc.py
```diff
@@ -547,13 +688,29 @@ def wait(self, timeout: float | None = None) -> Any:
...
-                                 *args: Any,
-                                 **kwargs: Any
-                                 ) -> Any:
+def rpc_property_call(
+    context: "qmi.core.context.QMI_Context",
+    rpc_object_address: QMI_MessageHandlerAddress,
+    property_name: str,
+    rpc_lock_token: QMI_LockTokenDescriptor | None,
+    property_value: Any = None,
+    *,
```
bvanommen, line 700: why use this star here so set_property_value cannot be positional?

qmi/core/rpc.py
```diff
@@ -694,9 +852,49 @@ def make_rpc_forward_function(method_name: str):
...
+                return
+
+            elif name == "_lock_token":
+                # Lock token should always be allowed to be set
+                object.__setattr__(self, name, value)
+
+            else:
+                raise AttributeError("Not allowed to set new attributes on a proxy class.")
+
+        object.__setattr__(self, name, value)
```
bvanommen, line 899: I'm not so sure when this fallback __setattr__ will get called. When the object is not initialized? Which situation would that be

qmi/core/rpc.py
```diff
@@ -1175,6 +1429,60 @@ def _handle_lock_rpc_request(self, request: QMI_LockRpcRequestMessage) -> QMI_Lo
...
+            # Modify the property; this can raise an exception or return a result.
+            try:
+                property = self._check_and_modify_property(request)
+                result_type = QMI_RpcFutureState.RESULT_IS_VALUE
+                result = property
+
+            except BaseException as exception:
+                _logger.debug("Modifying RPC property failed", exc_info=True)
+                result_type = QMI_RpcFutureState.RESULT_IS_EXCEPTION
+                result = exception
```
bvanommen, line 1451: I guess for consistency it would better to now send the traceback along in a tuple like we do for method calls

qmi/core/rpc.py
```diff
@@ -1175,6 +1429,60 @@ def _handle_lock_rpc_request(self, request: QMI_LockRpcRequestMessage) -> QMI_Lo
...
+        return reply
+
+    def _check_and_modify_property(self, request: QMI_PropertyRpcRequestMessage) -> Any:
+        """Check if the object has the property requested and is RPC-able; if so, return it."""
+        assert self._rpc_object is not None
+
+        # Check that the property was marked as RPC property.
+        if not hasattr(self._rpc_object, request.property_name):
+            raise QMI_UnknownRpcException(
+                f"Object {request.destination_address.object_id} of type {type(self._rpc_object).__name__}" +\
```
bvanommen, line 1474: Maybe the error message could explicitly state that it's not an RPC property? Likely a common error will be that you forget to set the property as RPC-able

bvanommen: On that note, you would somehow want to decorate class variables like we do methods. You could make a function RPC_Variable(var: Any) so that you declare variables as RPC_Variable(var_1) in the class. But I guess we'll never have that many RPC properties and they're all together at the start of the class declaration

qmi/core/util.py
```diff
@@ -82,6 +83,69 @@ def parse_address_and_port(address: str) -> tuple[str, int]:
...
+            return False
+        value2_signatures = [_make_value_structure_signature(item) for item in value2]
+        for item in value1:
+            signature = _make_value_structure_signature(item)
+            try:
+                value2_signatures.remove(signature)
+            except ValueError:
+                return False
+        return True
+
```
bvanommen, line 123: something like a python ndarray will not have its elements checked here, I think. Somehow I guess you want to make sure that any other structure that can be indexed an does not fall under the checked types here should return False instead of True

bvanommen: Ah I guess something like a string would be an exception, there length change is fine

bvanommen (changes requested):
That should be everything. I'll admit I didn't look at all the new tests in 100% detail. Most important is the tb stuff

qmi/core/rpc.py
heevasti: This should probably be handled like the other QMI_MessageDeliveryExceptions. I will change this to be the same.

qmi/core/rpc.py
heevasti: This one is the correct one.

qmi/core/rpc.py
heevasti: This `tuple` formatting is used now as a trick to detect if the received exception in the `wait` function is of old style exception without, or this new style exception with traceback data.

qmi/core/rpc.py
heevasti: I thought of that originally, but I want to have it clear that we will handle both cases, with the comment line after the (el)if.

qmi/core/rpc.py
heevasti: So, this error would come if an older QMI version tries to get a message/response from this newer QMI version RPC proxy, and it finishes in an exception. But not the other way around. I guess this is inevitable if we want to make this change, and not break the compatibility both ways? I didn't see a way to do it robustly both ways, at least.

qmi/core/rpc.py
heevasti: Just for future proofing that the any other calls (accidentally?) made to `rpc_property_call` will not start setting the property values without explicitly writing out the keyword input. Also some more reasoning in https://peps.python.org/pep-3102/.

qmi/core/rpc.py
heevasti: I think this relates to your comment in line 531. If we also add the traceback there, then the user should get the traceback if this fails. Otherwise, it would probably just fail silently.

qmi/core/rpc.py
heevasti: Yes. Although, that should not be the case probably ever... It's just more like a fallback now in case of irregular use.

qmi/core/rpc.py
heevasti: But here the error can be logged now if the `QMI_DEBUG` is set, together with `exc_info`. And this goes to the `_RpcThread` and not to the `QMI_RpcFuture` class.

qmi/core/rpc.py
heevasti: I will add in the text "RPC property" instead of simply "property". That is probably sufficient?

About the decoratable "RPC_Variable"s. Something in that direction is thought of in the project board under epics. Have a look ;) These "RPC properties" here are OK for these simpler class-level variables | attributes..

qmi/core/util.py
heevasti: Indeed, numpy arrays are not considered here element-wise. It is possible that someone might want to use a numpy array of course. Perhaps we can leave it for a possible future addition and not do it now?

qmi/core/rpc.py
heevasti: Sorry, that's not correct. It will be used until in the construction of the class the `self.rpc_nonblocking` has been set in L859. After that it should not get called.

PR #215, opened 2026-06-02
Qblox qmi driver

heevasti:
Add a QMI driver for Qblox Cluster series device. 

The driver contains only base functions to control the cluster, and trigger channels, and receive system info and module | function reference objects.

### DIFF
--- a/qmi/instruments/qblox/cluster.py
+++ b/qmi/instruments/qblox/cluster.py
@@ -1,0 +1,903 @@
+# mypy: disable-error-code=import-untyped
+# mypy: disable-error-code=import-not-found
+"""This module is for Qblox cluster control. It requires qblox_instruments 1.1.0 or newer."""
+
+from collections.abc import Callable
+from dataclasses import dataclass
+import inspect
+import logging
+import sys
+from functools import partial
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    import qblox_instruments
+    from qblox_instruments import Cluster
+    from qblox_instruments.native import Cluster as NativeCluster
+    from qblox_instruments.native.helpers import ChannelMapCache
+    from qblox_instruments.qcodes_drivers.module import Module as QcodesModule
+    from qblox_instruments.scpi.layers.cluster_mm_1_0 import Cluster as ScpiCluster
+    from qblox_instruments.types import DebugLevel
+
+else:
+    qblox_instruments = None
+    Cluster, NativeCluster, ScpiCluster = None, None, None
+    ChannelMapCache, QcodesModule, DebugLevel = None, None, None
+
+from qmi.core.context import QMI_Context
+from qmi.core.instrument import QMI_Instrument
+from qmi.core.rpc import rpc_method
+
+_logger = logging.getLogger(__name__)
+
+DEBUG_LEVEL: Any = None
+_EXTRA_ATTRS = [
+    "_get_sequencer_channel_map",
+    "_get_sequencer_acq_channel_map",
+    "_get_sequencer_state",
+    "_set_sequencer_channel_map",
+    "_set_sequencer_acq_channel_map",
+    "_set_acq_acquisition_data",
+    "_delete_awg_waveform",
+    "_delete_acq_weight",
+    "_delete_acq_acquisition",
+    "_delete_acq_acquisition_data",
+    "_add_acq_acquisition",
+    "_add_awg_waveform",
+    "_set_acq_acquisition_index",
+    "_set_awg_waveform_data",
+    "_set_awg_waveform_index",
+]
+
+# The amount of (external) trigger addresses in a cluster
+EXT_TRIGGERS_IN_CLUSTER = 15
+# Module configuration constants:
+# AO_IN_MODULE:              Analog output channels in modules with value range -1V...+1V.
+# AI_IN_MODULE:              Analog output channels in modules with value range -1V...+1V.
+# DIGITAL_MARKERS_IN_MODULE: Digital marker channels in modules, with values 'low' and 'high' (3.3V LVTTL).
+# SEQUENCERS_IN_MODULE:      Number of sequencers available in module.
+AO_IN_MODULE = {"QCM": 4, "QCM-RF": 2, "QRM": 2, "QRM-RF": 1, "QTM": 0, "MM": 0}
+AI_IN_MODULE = {"QCM": 0, "QCM-RF": 0, "QRM": 2, "QRM-RF": 1, "QTM": 0, "MM": 0}
+DIGITAL_MARKERS_IN_MODULE = {"QCM": 4, "QCM-RF": 2, "QRM": 4, "QRM-RF": 2, "QTM": 4, "MM": 0}
+SEQUENCERS_IN_MODULE = {"QCM": 6, "QCM-RF": 6, "QRM": 6, "QRM-RF": 6, "QTM": 8, "MM": 0}
+_get_required_qrm_qcm_attr_names: Callable[..., list[str]] | None = None
+_get_required_qtm_attr_names: Callable[..., list[str]] | None = None
+
+
+def _import_modules() -> None:
+    """Import the vendor-provided Qblox modules.
+    
+    This import is done in a function, instead of at the top-level,
+    to avoid an unnecessary dependency for programs that do not access
+    the instrument directly.
+    """
+    global qblox_instruments, Cluster, NativeCluster, ScpiCluster, ChannelMapCache, QcodesModule, DebugLevel  # noqa: PLW0603
+    global DEBUG_LEVEL, _get_required_qrm_qcm_attr_names, _get_required_qtm_attr_names  # noqa: PLW0603
+    if qblox_instruments is None:
+        import qblox_instruments
+        from qblox_instruments import Cluster
+        from qblox_instruments.native import Cluster as NativeCluster
+        from qblox_instruments.native.helpers import ChannelMapCache
+        from qblox_instruments.qcodes_drivers.module import Module as QcodesModule
+        from qblox_instruments.scpi.layers.cluster_mm_1_0 import Cluster as ScpiCluster
+        from qblox_instruments.types import DebugLevel
+
+        if tuple(map(int, qblox_instruments.__version__.split("."))) < tuple(map(int, "0.17.0".split("."))):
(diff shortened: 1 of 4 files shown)


qmi/instruments/qblox/cluster.py
```diff
@@ -0,0 +1,904 @@
...
+    "_set_acq_acquisition_index",
+    "_set_awg_waveform_data",
+    "_set_awg_waveform_index",
+]
+
+# The amount of (external) trigger addresses in a cluster
+EXT_TRIGGERS_IN_CLUSTER = 15
+# Module configuration constants:
+# AO_IN_MODULE:              Analog output channels in modules with value range -1V...+1V.
+# AI_IN_MODULE:              Analog output channels in modules with value range -1V...+1V.
```
bvanommen, line 56: input channels presumeably

qmi/instruments/qblox/cluster.py
```diff
@@ -0,0 +1,904 @@
...
+                getattr(self.cluster, f"_set_mrk_inv_en_{x}"), self.cluster
+            )
+            self.cluster_funcs[f"_get_mrk_inv_en_{x}"] = partial(
+                getattr(self.cluster, f"_get_mrk_inv_en_{x}"), self.cluster
+            )
+
+        # Add type handles as callables
+        for type_handle, value in self.cluster._type_handle.__dict__.items():
+            if type_handle.startswith("_is"):
+                # As 'value' gets overridden in memory for the lambda function, do like this:
```
bvanommen, line 448: overwritten

qmi/instruments/qblox/cluster.py
```diff
@@ -0,0 +1,904 @@
...
+            identifier=self.cluster_ip, port=self.port, debug=self.DEBUG_LEVEL, dummy_cfg=self.dummy_cfg
+        )
+        self._instrument_type = self.cluster.instrument_type.value  # Should return string "MM"
+        # dict {"<slot#>": {"<instrument_type>: _QbloxModule}, containing the modules.
+        # Startup module dictionary with MM; {"<slot#>": {"<instrument_type>: _QbloxModule},
+        # to contain all available modules.
+        idn = self.cluster._get_idn()
+        _, model, _, _ = idn.split(",")
+        self._modules = {"0": _QbloxModule(model, self.cluster_funcs)}
+        for k, v in self.cluster._mod_handles.items():
```
bvanommen, line 481: code would be a bit more readable if these keys and values have a bit more descriptive names

bvanommen: I guess the v's are modules?

qmi/instruments/qblox/cluster.py
```diff
@@ -0,0 +1,904 @@
...
+        # dict entry key="<slot#>", value={"<instrument_type>", func_refs": _QbloxModule}
+        for slot, module in self._modules.items():
+            # Then check if slot_no was given and matches. No need to go on if not.
+            if slot_no is not None and int(slot) != slot_no:
+                continue
+
+            if str(module) == module_type:
+                # Just double check module type before returning
+                if (
+                    module_type == "QCM"
```
bvanommen, line 563: if these two are on the same line it looks much better

qmi/instruments/qblox/cluster.py
```diff
@@ -0,0 +1,904 @@
...
+        self._add_cluster_funcs()
+        super().open()
+
+    @rpc_method
+    def reset_module(self, slot_no: int) -> None:
+        _logger.info(f"Resetting cluster module in slot {slot_no}.")
+        self.cluster._slot_reset()
+
+    @rpc_method
+    def get_module(self, module_type: str, slot_no: int | None = None) -> NativeCluster | _QbloxModule:
```
bvanommen, line 544: I guess it might be fair enough, but if you have multiple of the same module and you specify the module name, it's not clear which one you're gonna get. Maybe ideally you warn the user (or refuse) when you know there are duplicate modules and they didn't specify the slot.

qmi/instruments/qblox/cluster.py
```diff
@@ -0,0 +1,904 @@
...
+        module_type = module_type.upper()
+        sequencers: dict[str, Any] = {}
+        channels: dict[str, Any] = {}
+        module_func_refs = self.get_module_func_refs(module_type, slot_no)
+        # The channels definition in Qblox is done through the sequencers, and not the other way around. So, for our
+        # definition of control per channel we need to first map out the sequencers.
+        for sequencer in range(SEQUENCERS_IN_MODULE[module_type]):
+            try:
+                sequencer_config = module_func_refs["_get_sequencer_config"](sequencer)
+                if channel_type in [0, 1] and AI_IN_MODULE[module_type] > 0:
```
bvanommen, line 598: magical constants, ideally we replace

qmi/instruments/qblox/cluster.py
```diff
@@ -0,0 +1,904 @@
...
+                            channels[f"adc{channel}_IQ{sequencer}"] = sequencer_config["acq"]
+
+                    else:
+                        for channel in sequencer_channel_map[0]:
+                            channels[f"adc{channel}_acq_I{sequencer}"] = sequencer_config["awg"]
+
+                        for channel in sequencer_channel_map[1]:
+                            channels[f"adc{channel}_acq_Q{sequencer}"] = sequencer_config["awg"]
+
+                if channel_type in [0, 2] and AO_IN_MODULE[module_type] > 0:
```
bvanommen, line 612: same here

qmi/instruments/qblox/cluster.py
```diff
@@ -0,0 +1,904 @@
...
+                    sequencer_channel_map = module_func_refs["_get_sequencer_channel_map"](sequencer)
+                    # The first list defines 'I' type channels, the second 'Q' type
+                    for channel in sequencer_channel_map[0]:
+                        channels[f"dac{channel}_I{sequencer}"] = sequencer_config["awg"]  # TODO: Is this always valid?
+
+                    for channel in sequencer_channel_map[1]:
+                        channels[f"dac{channel}_Q{sequencer}"] = sequencer_config["awg"]  # TODO: Is this always valid?
+
+                if channel_type in [0, 3] and DIGITAL_MARKERS_IN_MODULE[module_type] > 0:
```
bvanommen, line 621: and here

qmi/instruments/qblox/cluster.py
```diff
@@ -0,0 +1,904 @@
...
+
+@dataclass()
+class TriggerThresholdConfig:
+    """Dataclass for holding marker channel trigger threshold parameters."""
+
+    threshold: Any
+    invert: Any
+
+
+class Qblox_QcodesCluster(Qblox_ClusterBase):
```
bvanommen, line 647: what is the difference between this class and `Qblox_NativeCluster`?

bvanommen (changes requested):
Tests look good, maybe bit of overlap between the two cluster modules.

I don't have experience with the device but looks OK. I would like to know why there are two classes. Is it to control either via SCPI commands or "native" control?

qmi/instruments/qblox/cluster.py
heevasti: The difference is that the 'NativeCluster' communicates directly with the cluster SCPI interface, while the 'QcodesCluster' communicates with the QCoDeS interface, which then communicates via SCPI interface (and a couple of special features). Both have their pros and cons, which I haven't managed to decide so far which one is overall better. The `qblox_instruments`  API has changed a lot over the development of various versions, and 'QcodesCluster' option was a bit more stable towards the changes, whereas 'NativeCluster' usually needed more smaller changes. On the other hand, at times the 'QcodesCluster' option's dependency 'Cluster' class radically changed with much more complex changes in the 'QcodesCluster' as result.

The other thing is that communicating directly with the SCPI interface we actually do not need the QCoDeS dependency. This reduces also considerably dependency conflicts with other packages as we can override or ignore the QCoDeS version be something else than in the `qblox_instruments` requirements. 

This is of course more maintenance work which is not desirable in the long term. But so far I couldn't weight one out against the other such that a decision could be made into which option we will focus on.

qmi/instruments/qblox/cluster.py
heevasti: Yes, will correct this, thanks!

qmi/instruments/qblox/cluster.py
heevasti: Will correct this, thanks!

qmi/instruments/qblox/cluster.py
heevasti: Yes, I will change 'k' to refer to the slot numbers and 'v' to module handles.

qmi/instruments/qblox/cluster.py
heevasti: The module you will get is the first match, i.e. with smallest slot number. I could add a message for the logger here to say the the first matching module will be 'get' when no slot number was provided. To 'know' if there are >1 modules of given type, I would have to change the loop to go to the end or do some pre-loop with the module names only. I think the call description (dosctring) in the base class and a logger info should be informative enough, without making extra checks here.

qmi/instruments/qblox/cluster.py
heevasti: I'm not sure if Ruff formatter will agree on that. I'll see what it thinks.

qmi/instruments/qblox/cluster.py
heevasti: Do you mean the `channel_type`? Those are explained in the base class docstring. But we could consider making enumerated integers class for them, which would make them more clear. Like
```python
class ChannelTypes(Enum.Enum):
    ALL_CHANNELS = 0  # all channels (default)
    AI_CHANNELS = 1  # analog input channels (adc)
    AO_CHANNELS = 2  # analog output channels (dac)
    MRK_CHANNELS =  3  # marker channels
    IO_CHANNELS = 4  # IO channels (QTM only)
```
and replace 
```python
if channel_type in [ChannelTypes.ALL_CHANNELS, ChannelTypes.AI] ...
```
Or perhaps with even shorter names as we use these only internally, really:
`_ChannelTypes.ALL, _ChannelTypes.AI`  etc.

qmi/instruments/qblox/cluster.py
heevasti: idem with previous

qmi/instruments/qblox/cluster.py
heevasti: idem with previous.

PR #217, opened 2026-06-25
Rpc traceback

### LINKED ISSUE #216: More elaborate error messages for RPC method calls
Description 
=========== 
RPC method calls will return the exception raised by the rpc method, however, debugging is often hard because the error message can be hard to interpret without further information. In fact, the traceback for the error disappears as it's caught in `_handle_method_rpc_request` , and is not sent to the caller. This has caused many users trouble in debugging.

On the host context, we can extract the traceback and send this, together with the error message, as a tuple back to the rpc caller. This is required because the `__traceback__` attribute of the exception class only exists inside the except block (i.e. it's not sent in the exception that is sent over with the rpc result).

Modules to be modified
---------------------- 
rpc.py

Tests to be created/updated 
--------------------------- 
test correct rpc reply in case of error

Documentation to be updated 
--------------------------- 
changelog.md

bvanommen:
Implementation of issue #216.

### DIFF
--- a/qmi/core/rpc.py
+++ b/qmi/core/rpc.py
@@ -508,7 +508,10 @@ class QMI_RpcFuture(QMI_MessageHandler):
                     # Check state of the future.
                     if self._state == QMI_RpcFutureState.RESULT_IS_EXCEPTION:
                         if not isinstance(self._result, tuple):
-                            raise QMI_RuntimeException("Received invalid exception object from RPC call.\
+if not isinstance(self._result, tuple):
+    if not isinstance(self._result, BaseException):
+        raise QMI_RuntimeException("Received invalid exception value from RPC call")
+    raise self._result
                             Is the RPC host running on an older QMI version?") from exc
                         exc = self._result[0]
                         tb = self._result[1]
--- a/qmi/core/rpc.py
+++ b/qmi/core/rpc.py
@@ -140,6 +140,7 @@ import inspect
 import logging
 import threading
 import time
+import traceback
 import enum
 from abc import ABCMeta
 from collections import deque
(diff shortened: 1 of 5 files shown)


heevasti (commented):
Looking good. Maybe add your name into the `CITATION.cff` file as well as you have contributed?

qmi/core/rpc.py
```diff
@@ -501,9 +507,20 @@ def wait(self, timeout: float | None = None) -> Any:
                 while True:
                     # Check state of the future.
                     if self._state == QMI_RpcFutureState.RESULT_IS_EXCEPTION:
-                        if not isinstance(self._result, BaseException):
-                            raise QMI_RuntimeException("Received invalid exception value from RPC call")
-                        raise self._result
+                        if not isinstance(self._result, tuple):
+                            raise QMI_RuntimeException("Received invalid exception object from RPC call.\
```
heevasti: In this case you should simply raise `self._result`, not to make a new exception.

heevasti:
```suggestion
if not isinstance(self._result, tuple):
    if not isinstance(self._result, BaseException):
        raise QMI_RuntimeException("Received invalid exception value from RPC call")
    raise self._result
```

qmi/data/dataset.py
```diff
@@ -792,7 +792,7 @@ def write_dataset_to_text(dataset: DataSet, fh: TextIO) -> None:
     if dataset.data.ndim == 1:
         rawdata = dataset.data.reshape(-1, 1)
     elif dataset.data.ndim > 2:
-        nrow = np.prod(dataset.data.shape[:-1])
+        nrow = int(np.prod(dataset.data.shape[:-1]))
         rawdata = dataset.data.reshape((nrow, dataset.ncol))
```
heevasti:
```suggestion
        rawdata = dataset.data.reshape((nrow, dataset.ncol))  # type: ignore
```

heevasti (approved):

